PR #12, opened 2021-05-11
Make it compile fibonacci

turlando:
- Add `PureScmTest.Fib` test module
- Add to AST: Identifier, Cond, Application and Lambda
- Extend Printer to emit Identifier, Cond, Application and Lambda
- Extend Transpiler to (sort of) handle Identifier, Cond, Application and Lambda
- Add simple optimizer to specialize sum and subtraction

Now we can compile Fibonacci, but the name resolution produces a wrong identifier for the recursive call.


f-f (approved):
I really like what you have going on here. Great work! 👏

PR #17, opened 2021-05-12
Handle multiple binders in case expression

### LINKED ISSUE #16: Add test for multiple binders in case expression
Compiling this

```
foo 0 0 = 0
foo 0 1 = 1
foo 1 0 = 10
foo _ 2 = 2
foo 3 _ = 30
foo _ _ = 50
```

will produce

```
(define foo
  (lambda (v)
    (lambda (v1)
      (cond
        ((= v 0) 0)
        ((= v1 0) 0)
        ((= v 0) 1)
        ((= v1 1) 1)
        ((= v 1) 10)
        ((= v1 0) 10)
        (? 2)
        ((= v1 2) 2)
        ((= v 3) 30)
        (? 30)
        (? 50)
        (? 50)))))
```

which is plain wrong.

turlando:
- Handle NullBinder
- Handle multiple binders

Now this code

```
foo 0 0 = 0
foo 0 1 = 1
foo 1 0 = 10
foo _ 2 = 2
foo 3 _ = 30
foo _ _ = 50
```

will translate into

```
(define foo
  (lambda (v)
    (lambda (v1)
      (cond
        ((and (= v 0) (= v1 0)) 0)
        ((and (= v 0) (= v1 1)) 1)
        ((and (= v 1) (= v1 0)) 10)
        ((and #t (= v1 2)) 2)
        ((and (= v 3) #t) 30)
        ((and #t #t) 50)))))
```

Maybe TODO (if it's the case make new cards):

- Optimize away `(and val #t)` into `val`
- Optimize away `(and #t #t)` into `#t` or `else`.

I think it's easier to handle those cases in the optimizer instead of emitting the optimized code straight away. What do you think?

This should close #16

### DIFF
--- a/resources/PureScmTest.FunctionMultipleMatch.corefn.json
+++ b/resources/PureScmTest.FunctionMultipleMatch.corefn.json
@@ -1,0 +1,532 @@
+{
+  "moduleName": [
+    "PureScmTest",
+    "FunctionMultipleMatch"
+  ],
+  "reExports": {
+  },
+  "imports": [
+    {
+      "annotation": {
+        "meta": null,
+        "sourceSpan": {
+          "start": [
+            1,
+            1
+          ],
+          "end": [
+            8,
+            13
+          ]
+        }
+      },
+      "moduleName": [
+        "Prim"
+      ]
+    }
+  ],
+  "builtWith": "0.14.1",
+  "modulePath": "src/PureScmTest/FunctionMultipleMatch.purs",
+  "exports": [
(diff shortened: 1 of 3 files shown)


f-f:
> I think it's easier to handle those cases in the optimizer instead of emitting the optimized code straight away. What do you think?

Totally. Also it could likely be that Chez will optimize this away, I wouldn't be surprised

f-f (approved):

PR #18, opened 2021-05-12
Handle VarBinder

turlando:


src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -108,18 +109,27 @@ unifyBinder :: Expr Ann -> Binder Ann -> AST
...
+unifyBinder _value (VarBinder _ann _ident) = Identifier "#t"
 
 unifyBinder _ _ = error "Not implemented"
 
 
+valuesAndBindersAndResultToLet :: [(Expr Ann, Binder a)] -> Expr Ann -> AST
+valuesAndBindersAndResultToLet valuesAndBinders expr =
+  let
+    go (value, (VarBinder _ann ident)) = (runIdent ident, exprToScheme value)
+    go (_, _) = ("#t", Identifier "#t")
```
turlando, line 128: This is the tricky bit.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -88,16 +88,17 @@ caseAlternativesToScheme values caseAlternatives =
     condClauses = map makeClause alternatives
 
     makeClause (_valuesAndBinders, Left _xs) = error "Not implemented"
-    makeClause (valuesAndBinders, Right expr) = 
-      (makeTest valuesAndBinders, exprToScheme expr)
+    makeClause (valuesAndBinders, Right result) = 
+      (makeTest valuesAndBinders,
+       valuesAndBindersAndResultToLet valuesAndBinders result)
```
turlando, line 93: The first element of the tuple is the test and the second argument is the result.

turlando:
I think I got something working even if not ideal yet.

```
(define foo
  (lambda (v)
    (lambda (v1)
      (cond
        ((and (= v 0) (= v1 0)) (let () 0))
        ((and (= v 0) (= v1 1)) (let () 1))
        ((and (= v 1) (= v1 0)) (let () 10))
        ((and #t (= v1 2)) (let () 2))
        ((and (= v 3) #t) (let () 30))
        ((and (= v 4) #t) (let ((n v1)) n))
        ((and #t (= v1 5)) (let ((m v)) m))
        ((and #t #t) (let ((m v) (n v1)) 100))))))
```

turlando:
I think this should also close #16

f-f (approved):
:clap:

PR #20, opened 2021-05-14
Add end to end tests

turlando:


test/EndToEndSpec.hs
```diff
@@ -0,0 +1,38 @@
...
+
+compile :: FilePath -> IO Text
+compile path = (printScheme . runOptimizations . moduleToScheme)
+           <$> readJSONFile path
+
+spec :: Spec
+spec = do
+  it "parse Literals" $ do
+    res <- compile "test/purs/PureScmTest.Literals.corefn.json"
+    res `shouldBe` "(define integerLiteral2 42)\n\n(define integerLiteral1 23)\n\n(define integerArray2 (vector 23 42))\n\n(define integerArray1 (vector 1 2 3))"
```
f-f, line 30: We should store these on files - we could have a `test/fixtures` folder where we store these, and read them in with [`readTextFile`](https://www.stackage.org/haddock/lts-17.11/turtle-1.5.20/Turtle-Prelude.html#v:readTextFile) from `turtle`.

Of course it would be nice to define a `readFixture` that reads from the right fixture folder so you don't always have to specify the prefix path

test/EndToEndSpec.hs
```diff
@@ -0,0 +1,38 @@
...
+import qualified Data.Aeson.Types                              as JSON.T
+import qualified Language.PureScript.CoreFn.FromJSON           as CoreFn
+import           Language.PureScript.CoreFn.Module             (Module)
+import           Language.PureScript.CoreFn.Ann                (Ann)
+import           Language.PureScript.Scheme.CodeGen.Transpiler (moduleToScheme)
+import           Language.PureScript.Scheme.CodeGen.Printer    (printScheme)
+import           Language.PureScript.Scheme.CodeGen.Optimizer  (runOptimizations)
+
+readJSONFile :: FilePath -> IO (Module Ann)
+readJSONFile path = do
```
f-f, line 15: This is duplicated code - you should share it. Also the name is not that accurate, since it's more of a `readModuleFromJson`

test/EndToEndSpec.hs
```diff
@@ -0,0 +1,38 @@
...
+  case res of
+    Nothing -> error "Nope"
+    Just (_version, module_) -> return module_
+
+compile :: FilePath -> IO Text
+compile path = (printScheme . runOptimizations . moduleToScheme)
+           <$> readJSONFile path
+
+spec :: Spec
+spec = do
```
f-f, line 27: Instead of versioning the corefn files, we should just have a PureScript project (somewhere like `test/test-project`) and compile it before running the tests. Something like:
```suggestion
setup = do
  oldDir <- Turtle.pwd
  Turtle.cd "test/test-project"
  (exitCode, out, err) <- Turtle.shellStrictWithErr "spago build -u '-g corefn'"
  Turtle.cd oldDir
  case exitCode of
    ExitSuccess -> pure ()
    ExitFailure -> error "spago failed to run"

spec = beforeAll_ setup $ do
```

Then you could just read the resulting corefn files from `output`

f-f (commented):
This is a great starting point! I left some comments to make the tests a little leaner - I implied we should pull in [`turtle`](https://www.stackage.org/haddock/lts-17.11/turtle-1.5.20/index.html) as a dependency, since it would make it easy to do commandline stuff.

Other than that we'll have to:
- run tests in CI, i.e. add a `stack test` step to the workflow file
- ..and for that to work we'll need to install `purs` and `spago` in CI before running the tests. You can just use the [`setup-purescript`](https://github.com/purescript-contrib/setup-purescript) action for that, just follow the tutorial in the readme and everything should just work

turlando:
I refactored the tests.

For simplicity for now I'm not creating a test PureScript project to run the tests against. I will do it when we will emit proper Scheme libraries instead of top-level code.

I managed to run the tests without importing turtle (for now).

Makefile
```diff
@@ -0,0 +1,5 @@
+build-corefn:
```
turlando, line 1:
```
.PHONY: build-corefn
build-corefn: ...
```

Even better: just `corefn` instead of `build-corefn` since you're already `make`ing it.

Sample applies for `regen-scheme`.

Also the file is missing a newline at the bottom.

tools/RegenScheme.hs
```diff
@@ -0,0 +1,20 @@
+module RegenScheme where
```
turlando, line 1: This module is always generating the scheme files, so why it's called Regen?

tools/RegenScheme.hs
```diff
@@ -0,0 +1,20 @@
...
+
+import qualified Turtle
+import qualified Language.PureScript.Scheme.Make as PureScm
+import Data.Foldable (for_)
+
+modules :: [String]
+modules =
```
turlando, line 8: Dupe

test/Language/PureScript/Scheme/MakeSpec.hs
```diff
@@ -1,29 +1,37 @@
...
 testModule module_ = do
   it ("Can compile " <> module_) $ do
     compilationResult <- compile $ coreFnPath module_
     expectedResult <- T.IO.readFile $ schemePath module_
     compilationResult `shouldBe` expectedResult
-  
+
+setup :: IO ()
+setup = do
+  oldDir <- Turtle.pwd
```
turlando, line 28: What about something like (maybe to have in a separate file?)

```
inDir :: Turtle.MonadIO m => Turtle.FilePath -> m b -> m b
inDir directory do_ = do
  oldDir <- Turtle.pwd
  Turtle.cd directory
  res <- do_
  Turtle.cd oldDir
  return res

runInDir
  :: Turtle.MonadIO m
  => Turtle.FilePath
  -> Turtle.Text
  -> m (Turtle.ExitCode, Turtle.Text, Turtle.Text)
runInDir directory cmd = inDir directory $ Turtle.shellStrictWithErr cmd mempty
```

and then

```
setup :: IO ()
setup = do
  (exitCode, _out, err) <- runInDir "test/test-project" "spago build -u '-g corefn'"
  case exitCode of
    Turtle.ExitSuccess -> pure ()
    Turtle.ExitFailure _ -> error $ "spago failed to run. error: " <> show err
```

It would also be nice to have the constants factored out.

turlando (commented):
`shell.nix` it's missing `spago` in the environment and tests will fail without it. Also maybe we should think in the future to use Nix in the CI to have an unified test and development environment.

turlando:
The suggested patch does not work on my system:

```
nix-shell:~/purescm]$ stack test
purescm> test (suite: spec)


Language.PureScript.Scheme.Make
  Can compile PureScmTest.Literals FAILED [1]
  Can compile PureScmTest.Fib
    # PENDING: exception in beforeAll-hook (see previous failure)
  Can compile PureScmTest.FunctionMultipleMatch
    # PENDING: exception in beforeAll-hook (see previous failure)

Failures:

  test/Language/PureScript/Scheme/MakeSpec.hs:45:29: 
  1) Language.PureScript.Scheme.Make Can compile PureScmTest.Literals
       uncaught exception: ErrorCall
       spago error: "/bin/sh: spago: command not found\n"
       CallStack (from HasCallStack):
         error, called at test/Language/PureScript/Scheme/MakeSpec.hs:45:29 in main:Language.PureScript.Scheme.MakeSpec

  To rerun use: --match "/Language.PureScript.Scheme.Make/Can compile PureScmTest.Literals/"

Randomized with seed 1196109945

Finished in 0.0069 seconds
3 examples, 1 failure, 2 pending

purescm> Test suite spec failed
Test suite failure for package purescm-0.0.1
    spec:  exited with: ExitFailure 1
Logs printed to console


[nix-shell:~/purescm]$ spago version
0.16.0
```

I have no time to fix this now. Moving these (useful) patches to the `test-with-project` branch and I will work on that when I will have time.

Merging the previous (working) patches and creating a card for making the test run off a real project.

PR #24, opened 2021-05-20
Handle top-level recursive binds

### LINKED ISSUE #22: Handle top-level recursive binds
Right now we're translating recursive top-level binds (`Rec [((a, Ident), Expr a)] deriving (Show, Functor)`) by picking the first bind and ignoring the following ones.

This works fine for simple recursive functions such as fibonacci, but spectacularly fails for mutually recursive functions and much more.

turlando:
This should close #22


f-f (approved):
👏

PR #25, opened 2021-05-20
Handle guarded CaseAlternatives

### LINKED ISSUE #23: Handle guarded CaseAlternatives

turlando:
Should close #23 and make Eris happy.


test/resources/PureScmTest.GuardedFunction.sls
```diff
@@ -0,0 +1,3 @@
+(define min (lambda (n) (lambda (m) (cond ((((Data.Ord.lessThan Data.Ord.ordInt) n1) m1) n) (Data.Boolean.otherwise m)))))
+
+(define gcd (lambda (v) (lambda (v1) (cond ((= v1 0) v) ((= v 0) v1) ((((Data.Ord.greaterThan Data.Ord.ordInt) n) m) ((PureScmTest.GuardedFunction.gcd (- n m)) m)) (Data.Boolean.otherwise ((PureScmTest.GuardedFunction.gcd n) (- m n)))))))
```
f-f, line 3: It looks like the body is referencing `n` and `m`, but they are not bound.

We should start having tests where we actually run the Scheme that is generated, so we know that it works as intended

turlando: That was a huge error on my side. Should be fixed now.

f-f (approved):
Looks great now! 👏

(But let's prioritize those functional tests)

PR #26, opened 2021-05-24
Handle Constructor and ConstructorBinder

turlando:


src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -107,23 +120,31 @@ caseToScheme values caseAlternatives =
...
+    -- TODO: Optimize (and foo) into foo.
     test valueAndBinders =
-      Application (Identifier "and")
-                  (map (\(value, binder) -> binderToTest value binder)
-                       valueAndBinders)
+      and_ (concatMap (\(value, binder) -> binderToTest (exprToScheme value)
+                                                        binder)
+           valueAndBinders)
 
     -- Emit a cond clause test for each value and binder.
```
turlando, line 135: Update comment.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -107,23 +120,31 @@ caseToScheme values caseAlternatives =
...
-      Application (Identifier "=") [exprToScheme value, IntegerLiteral integer]
-    binderToTest _value (VarBinder _ann _ident) = Identifier "#t"
-    binderToTest _value (NullBinder _ann) = Identifier "#t"
+      [eq [value, IntegerLiteral integer]]
+    binderToTest _value (VarBinder _ann _ident) = [t]
+    binderToTest value (ConstructorBinder _ann _typeName constructorName binders) =
+      (:) (eqQ (car value)
+               (quote (Identifier (runProperName (disqualify constructorName)))))
+          (concatMap (\(i, b) -> (binderToTest (vectorRef value (IntegerLiteral i)) b))
+                     (zip [1 .. (genericLength binders)]
```
turlando, line 146: Off by one.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -140,8 +161,23 @@ caseToScheme values caseAlternatives =
...
-      | (value, (VarBinder _ann ident)) <- valuesAndBinders ]
+      concatMap go valuesAndBinders
+      where
+        go (_value, NullBinder _ann) = []
+        go (_value, LiteralBinder _ann _literal) = []
+        go (value, VarBinder _ann ident) = [(runIdent ident, exprToScheme value)]
+        go (value, ConstructorBinder _ann _typeName _constructorName binders) =
+          map (\(i, b) -> (go' b, vectorRef (exprToScheme value)
+                                            (IntegerLiteral i)))
+              (zip [1 .. (genericLength binders)]
```
turlando, line 172: Off by one.

turlando:
In `(and (eq? (car v) (quote Bar)) (= (vector-ref v 1) 1)`

- `(eq? (car v) (quote Bar))` is correct because Bar is a symbol and `eq?` performs address comparison.
- `(= (vector-ref v 1) 1)` is NOT correct because `=` can only compare numbers. Should we just use `equal?` and forget about it?

---

From https://stackoverflow.com/a/17719745

- Use the = predicate when you wish to test whether two numbers are equivalent.
- Use the eqv? predicate when you wish to test whether two non-numeric values are equivalent.
- Use the equal? predicate when you wish to test whether two lists, vectors, etc. are equivalent.=
- Don't use the eq? predicate unless you know exactly what you're doing.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -30,6 +36,13 @@ exprToScheme :: Expr Ann -> AST
 exprToScheme (Literal _ann literal) =
   literalToScheme literal
 
+exprToScheme (Constructor _ann _typeName constructorName fields) =
+  go fields
+  where
+    go (x:xs) = Lambda (runIdent x) (go xs)
+    go [] = cons (quote (Identifier (runProperName constructorName)))
+                 (vector (fmap (\field -> Identifier (runIdent field)) fields))
```
f-f, line 44: Why do we turn a constructor into a lambda here? It would be good to have a comment explaining what's going on here

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -107,23 +120,31 @@ caseToScheme values caseAlternatives =
...
-    -- If we're handling multiple values and binders for a single result
-    -- we have to wrap the various resulting test into an `and' expression.
+    -- Emit the test for a cond clause.
     -- In scheme: (cond ((and test1a test1b) result1) ...)
-    test [(value, binder)] = binderToTest value binder
+    -- Always wrap the test in an `and' form since binderToTest has to return a
+    -- list because when applied to a ConstructorBinder it returns a list of AST.
+    -- We could have generated an `and' expression in such case in binderToTest,
+    -- but it's easier to optimize (and foo) -> foo.
+    -- TODO: Optimize (and foo) into foo.
```
f-f, line 129: I thought this was already happening?

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -140,8 +161,23 @@ caseToScheme values caseAlternatives =
...
+                                            (IntegerLiteral i)))
+              (zip [1 .. (genericLength binders)]
+                   binders)
+        go _ = error "Not implemented"
+
+        -- TODO: possible fire hazard. So far I've met no ConstructorBinder
+        -- whose binders are anything but VarBinders. But if that doesn't hold
+        -- then it will explode here.
+        go' (VarBinder _ann ident) = runIdent ident
+        go' _ = error "Not implemented"
```
f-f, line 180: Does this break it?

```
data Foo = Foo String

data Bar = Bar Foo

match :: Bar -> Int
match (Bar (Foo "")) = 1
match _ = 23
```

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -140,8 +161,23 @@ caseToScheme values caseAlternatives =
...
+      where
+        go (_value, NullBinder _ann) = []
+        go (_value, LiteralBinder _ann _literal) = []
+        go (value, VarBinder _ann ident) = [(runIdent ident, exprToScheme value)]
+        go (value, ConstructorBinder _ann _typeName _constructorName binders) =
+          map (\(i, b) -> (go' b, vectorRef (exprToScheme value)
+                                            (IntegerLiteral i)))
+              (zip [1 .. (genericLength binders)]
+                   binders)
+        go _ = error "Not implemented"
```
f-f, line 174: Can we explicitly match on all cases?

test/resources/PureScmTest.Constructor.purs
```diff
@@ -0,0 +1,10 @@
+module PureScmTest.Constructor where
+
+data Foo = Bar Int Int
+         | Baz
+
+match (Bar 1 2)   = 0
+match (Bar 1 _)   = 1
+match (Bar _ 2)   = 2
+match (Bar i1 i2) = i1
+match Baz = 0
```
f-f, line 10: Is this test exercising all the code paths?

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
f-f: Totally, also it looks like you're using this in multiple places, so it would be safer to have a helper, something like
```haskell
mapIndexed :: (a -> Int -> b) -> [a] -> [b]
mapIndexed f l = zipWith f l [0..]
```
Note that you don't need to specify the length - Haskell has lazy lists, and the `zip` will take values until your collection ends. Magic :slightly_smiling_face:

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -107,23 +120,31 @@ caseToScheme values caseAlternatives =
...
     binderToTest value (LiteralBinder _ann (NumericLiteral (Left integer))) =
-      Application (Identifier "=") [exprToScheme value, IntegerLiteral integer]
-    binderToTest _value (VarBinder _ann _ident) = Identifier "#t"
-    binderToTest _value (NullBinder _ann) = Identifier "#t"
+      [eq [value, IntegerLiteral integer]]
+    binderToTest _value (VarBinder _ann _ident) = [t]
+    binderToTest value (ConstructorBinder _ann _typeName constructorName binders) =
+      (:) (eqQ (car value)
+               (quote (Identifier (runProperName (disqualify constructorName)))))
+          (concatMap (\(i, b) -> (binderToTest (vectorRef value (IntegerLiteral i)) b))
```
f-f, line 145: I read this code a couple of times and I still have no idea what's going on here, so a comment with the reasoning would really help

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: Uhm... "this" what?

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: I can't really tell because strings are not implemented yet.

But I can test this code (basically the same swapping String with Int):

```
data Foo = Foo Int

data Bar = Bar Foo

match :: Bar -> Int
match (Bar (Foo 0)) = 1
match _ = 23
```

And it results into:

```
(define Foo
  (lambda (value0)
    (cons (quote Foo)
              (vector value0))))

(define Bar
  (lambda (value0)
    (cons (quote Bar)
              (vector value0))))

(define match
  (lambda (v)
    (cond ((and (eq? (car v) (quote Bar))
                (eq? (car (vector-ref v 1)) (quote Foo))
                (= (vector-ref (vector-ref v 1) 1) 0)) 1)
          (#t 23))))
```

It looks correct to me, the only bug being the off-by-one error we already know about.

test/resources/PureScmTest.Constructor.purs
turlando: In theory yes: I ran `stack test --coverage` and `stack hpc report .` and it says that the only never executed paths are the `error` placeholders.

In practice: we should test nested constructor even though I tested your previous test case and it seems to work.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: Just found another bug: we should run `vector-ref` against the `cdr` of `v` and not `v` itself.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
f-f: So like `var@(Bar i1 i2)`? If that's so, can you add this to the test? (also referencing `var` in the body so we make sure to handle the renaming)

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
f-f: It's definitely time to run this generated Scheme files

test/resources/PureScmTest.Constructor.purs
f-f: Let's add nested constructors to this test, and let's try to hit all the `error` placeholders

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -30,6 +36,38 @@ exprToScheme :: Expr Ann -> AST
 exprToScheme (Literal _ann literal) =
   literalToScheme literal
 
+-- An ADT is translated to a tagged pair. The car of the pair is the tag,
+-- holoding the name of the Constructor as a quoted symbol. The cdr of the
```
f-f, line 40:
```suggestion
-- holding the name of the Constructor as a quoted symbol. The cdr of the
```

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -30,6 +36,38 @@ exprToScheme :: Expr Ann -> AST
...
+-- A product type with N fields results into a curried function of N arguments.
+--     data Foo = Bar Int Int
+--     (lambda (v0) (lambda (v1) (cons 'Bar (vector v0 v1)))) ; defined as Bar
+--
+-- A sum type results into multiple Constructor. If the sum type is `simple'
+-- we just emit a constant representing that constructed value.
+--     data Foo = Bar | Baz
+--     (cons (quote Bar) (vector)) ; defined as Bar
+--     (cons (quote Baz) (vector)) ; defined as Baz
+-- TODO: maybe we could remove that empty vector?
```
f-f, line 56: Yeah you could literally put `nil` in there, since you're never going to access this field anyways

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -30,6 +36,38 @@ exprToScheme :: Expr Ann -> AST
...
+--
+-- If the sum type is not simple, so its constructors could be product types
+-- we amit multiple functions.
```
f-f, line 59:
```suggestion
-- we emit multiple functions.
```

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
f-f: Optimizing the `and` away

turlando: We actually do and I forgot lmao

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: Let's open a new card

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: Is it clearer now?

test/resources/PureScmTest.Constructor.purs
turlando: I will add another test case (NestedConstructor.purs) with the following code

```
data Foo = Foo Int

data Bar = Bar Foo

match :: Bar -> Int
match (Bar (Foo 0)) = 1
match _ = 23
```

Let me know if I have to change anything.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: The wollowing test case

```
data Foo = Bar Int Int
         | Baz

match var@(Bar i1 i2) = var
match Baz = Baz
```

hits `binderToTest` with a NamedBinder. I'd rather implement it in the next PR since there is already a lot going on in this one. Let me know if I have to test with a different case though.

turlando:
Something we haven't discussed yet. As of now we're using the constructor name only as tag for the dotted pair. Should we use a fully qualified name instead?

If so the issue is that while at the binder level we've got access to the namespace, since it's a `ConstructorBinder a (Qualified (ProperName 'TypeName)) (Qualified (ProperName 'ConstructorName)) [Binder a]`, at the constructor level we don't because we just have a `Constructor a (ProperName 'TypeName) (ProperName 'ConstructorName) [Ident]`.

What do we do here?

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: True but should we do it in this PR since it's getting pretty big already?

Also: should we do it at all? It's more logic... for what?

turlando: Addendum: if we want to do this, we might want to just have `'Bar` and `'Baz` instead of `('Bar . #())` or `('Bar . nil)`. It's definitely more lisp-y. But in any case I'd rather do this in a different PR.

f-f: Sounds good.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
f-f: Yep, this test looks good, and fixing this in the next PR is a good idea too

test/resources/PureScmTest.Constructor.purs
f-f: Perfect!

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
f-f: Possibly? :smile: 
The comment really helps, so I'd say yes.

f-f:
> What do we do here?

This is going to sound redundant, but...what's the JS backend doing?

turlando:
This

```
data Foo = Bar Int Int
         | Baz

derive instance eqFoo :: Eq Foo

match (Bar i1 i2) = i1
match Baz = 0

same :: Foo -> Foo -> Boolean
same x y = x == y
```

gets translated into

```
var Data_Eq = require("../Data.Eq/index.js");

var Bar = (function () {
    function Bar(value0, value1) {
        this.value0 = value0;
        this.value1 = value1;
    };
    Bar.create = function (value0) {
        return function (value1) {
            return new Bar(value0, value1);
        };
    };
    return Bar;
})();

var Baz = (function () {
    function Baz() {

    };
    Baz.value = new Baz();
    return Baz;
})();

var match = function (v) {
    if (v instanceof Bar) {
        return v.value0;
    };
    if (v instanceof Baz) {
        return 0;
    };
    throw new Error("Failed pattern match at PureScmTest.Constructor (line 10, column 1 - line 10, column 23): " + [ v.constructor.name ]);
};

var eqFoo = new Data_Eq.Eq(function (x) {
    return function (y) {
        if (x instanceof Bar && y instanceof Bar) {
            return x.value0 === y.value0 && x.value1 === y.value1;
        };
        if (x instanceof Baz && y instanceof Baz) {
            return true;
        };
        return false;
    };
});

var same = function (x) {
    return function (y) {
        return Data_Eq.eq(eqFoo)(x)(y);
    };
};
```

The type comparison is done using `instanceof`, which should rule out disambiguation.

My concern is the following: what would happen if I have Foo from module A and Foo from module B? Technically the compiler should not allow type mismatch at compile time, but what about interop?

test/resources/PureScmTest.Constructor.purs
turlando: The JS compiler translates that code into

```
var Foo = (function () {
    function Foo(value0) {
        this.value0 = value0;
    };
    Foo.create = function (value0) {
        return new Foo(value0);
    };
    return Foo;
})();

var Bar = (function () {
    function Bar(value0) {
        this.value0 = value0;
    };
    Bar.create = function (value0) {
        return new Bar(value0);
    };
    return Bar;
})();

var match = function (v) {
    if (v.value0.value0 === 0) {
        return 1;
    };
    return 23;
};
```

while the Scheme compiler does this

```
(define Foo
  (lambda (value0)
    (cons (quote Foo) (vector value0))))

(define Bar
  (lambda (value0)
    (cons (quote Bar) (vector value0))))

(define match
  (lambda (v)
    (cond ((and (eq? (car v) (quote Bar))
                (eq? (car (vector-ref (cdr v) 0)) (quote Foo))
                (= (vector-ref (cdr (vector-ref (cdr v) 0)) 0) 0)) 1)
          (#t 23))))
```

Are we doing way too many tests?

turlando: Discussion about this moved to #27

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: Issue #29 will take care of this

f-f (approved):

Issue #29, opened 2021-05-26
Handle NamedBinder


turlando:

turlando:
Previous discussion [here](https://github.com/turlando/purescm/pull/26#discussion_r638698117).

Issue #27, opened 2021-05-26
Don't emit constructor checks when dealing with types with a single constructor


turlando:
When we have
`data Foo = Foo`
we translate it into
`(cons 'Foo . (vector))`

When we have
`f Foo = 23`
we always check if `(car v)` is `'Foo` but it always will be because the type checker will not allow otherwise.

Cfr. https://github.com/turlando/purescm/pull/26#discussion_r639454003

turlando:
Discussion continues from [here](https://github.com/turlando/purescm/issues/30#issuecomment-850787472)

f-f:
We need to verify how the new implementation handles this, and if we need to do anything at all

turlando:
Since #97, the following code

```purescript
data NoArgs = NoArgs

test1 :: NoArgs -> Boolean
test1 = case _ of
  NoArgs -> true

foo :: Boolean
foo = test1 NoArgs
```

compiles into

```scheme
(scm:define NoArgs
  (scm:quote NoArgs))

(scm:define NoArgs?
  (scm:lambda (v)
    (scm:eq? (scm:quote NoArgs) v)))

(scm:define test1
  (scm:lambda (v0)
    #t))

(scm:define foo
  #t)
```

So I think we can close this issue?

PR #31, opened 2021-05-26
Handle NamedBinder

turlando:
Should solve #29

### DIFF
--- a/src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
+++ b/src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
@@ -208,12 +208,19 @@ caseToScheme values caseAlternatives =
            (\i b -> (binderToTest (vectorRef (cdr value) (IntegerLiteral i)) b))
            binders)
 
-    -- We have a level of indirection. `value' is bound to `binder' through
-    -- `ident'. For instance in
+    -- Consider the case:
     --     data Foo = Bar Int
     --     match bar@(Bar i) = bar
-    -- `value' is an Identifier referring to (Bar i), ident is an Ident "bar"
-    -- and binder is the ConstructorBinder for (Bar i).
+    -- It get translated to:
+    --    (define Bar (lambda (x) ('Bar . x)))
+    --    (define match (lambda (v) (cond (eq? (car v) 'Bar) v)))
+    -- `bar' (in PureScript) is an alias to `v' (in Scheme). `value' is an
+    -- `Expr.Var "v"', and `ident' is an `Ident.Ident "bar"'.
+    -- So in order to generate the test for the cond in the case of a NamedBinder
+    -- we have to process its binder (with binderToTest) using the actual
+    -- variable name `value' and not with the alias `ìdent'.
+    -- We will have to use ident when we have to deal with the result associated
+    -- to this case though.
     binderToTest value (NamedBinder _ann _ident binder) = binderToTest value binder
 
     -- Emit a cond clause result for multiple values and binders associated to
@@ -247,10 +254,17 @@ caseToScheme values caseAlternatives =
           mapWithIndex (\i b -> (go' b, vectorRef (cdr value) (IntegerLiteral i)))
                        binders
 
-        -- We have to unfold the level of indirection introduced by the
-        -- NamedBinder. We do so by replacing its ident with its value.
-        -- We also have to handle the binder associated to the NamedBinder.
-        -- We do so by recursively calling go against the binder.
+
+        -- Consider the case:
+        --     data Foo = Bar Int
+        --     match bar@(Bar i) = bar
+        -- It get translated to:
+        --    (define Bar (lambda (x) ('Bar . x)))
+        --    (define match (lambda (v) (cond (eq? (car v) 'Bar) v)))
+        -- `value' is an `Expr.Var "v"', and `ident' is an `Ident.Ident "bar"'.
+        -- We have to do multiple replacements here. First, we have to replace
+        -- each each `ident' with `value'. Then, we have to apply other possible
+        -- replacements for the binder associated to the NamedBinder.
         go value (NamedBinder _ann ident binder) =
           (:) (runIdent ident, value)
               (go value binder)
--- a/src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
+++ b/src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
@@ -230,16 +230,16 @@ caseToScheme values caseAlternatives =
     -- We have to replace each occurrence of `m' with `v' and each occurrence
     -- of `n' with `v0'.
     variablesToReplace valuesAndBinders =
-      concatMap go valuesAndBinders
+      concatMap (\(value, binder) -> go value binder) valuesAndBinders
       where
-        go (_value, NullBinder _ann) = []
-        go (_value, LiteralBinder _ann _literal) = []
-        go (value, VarBinder _ann ident) = [(runIdent ident, exprToScheme value)]
+        go _value (NullBinder _ann) = []
+        go _value (LiteralBinder _ann _literal) = []
+        go value (VarBinder _ann ident) = [(runIdent ident, exprToScheme value)]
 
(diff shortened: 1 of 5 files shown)


src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -206,7 +207,14 @@ caseToScheme values caseAlternatives =
...
            binders)
-    binderToTest _ _ = error "Not implemented"
+
+    -- We have a level of indirection. `value' is bound to `binder' through
+    -- `ident'. For instance in
+    --     data Foo = Bar Int
+    --     match bar@(Bar i) = bar
+    -- `value' is an Identifier referring to (Bar i), ident is an Ident "bar"
+    -- and binder is the ConstructorBinder for (Bar i).
+    binderToTest value (NamedBinder _ann _ident binder) = binderToTest value binder
```
f-f, line 217: I am confused about why we throw away identifiers here and in other places.
Like, what happens when we have something like this? (this might be a useful test to include anyways)
```
data Foo = Bar Int

doubleFoo :: Foo -> Int
doubleFoo bar@(Bar i) = case bar of
  Bar j -> i + j
```

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -216,27 +224,36 @@ caseToScheme values caseAlternatives =
...
-    -- Each identifier in a VarBinder will be replaced by its corresponding
-    -- value. E.g. `foo m n = m + n' will emit a function whose parameters are
-    -- v and v0, but the `m + n' expression will use `m' and `n' as variable.
-    -- We have to replace each occurrence of `m' with `v' and each occurrence
-    -- of `n' with `v0'.
+
+    variablesToReplace :: [(Expr Ann, Binder a)] -> [(Text, AST)]
     variablesToReplace valuesAndBinders =
-      concatMap go valuesAndBinders
+      concatMap (\(value, binder) -> go (exprToScheme value) binder)
```
f-f, line 230: Why this `exprToScheme`? This probably deserves a comment

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -216,27 +224,36 @@ caseToScheme values caseAlternatives =
...
         -- of the tagged pair holding the Constructor values.
-        go (value, ConstructorBinder _ann _typeName _constructorName binders) =
-          mapWithIndex (\i b -> (go' b, vectorRef (cdr (exprToScheme value))
-                                                  (IntegerLiteral i)))
+        go value (ConstructorBinder _ann _typeName _constructorName binders) =
+          mapWithIndex (\i b -> (go' b, vectorRef (cdr value) (IntegerLiteral i)))
                        binders
 
-        go _ = error "Not implemented"
+        -- We have to unfold the level of indirection introduced by the
```
f-f, line 250: I am not sure what you mean by "level of indirection" here and above

f-f (commented):
Good work :clap: 
The implementation clearly works, but I feel like I need more comments to understand this code :slightly_smiling_face:

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando:
```
(define Bar
  (lambda (value0)
    (cons (quote Bar) (vector value0))))

(define doubleFoo
  (lambda (v)
    (cond ((eq? (car v) (quote Bar))
	   (cond ((eq? (car v) (quote Bar))
		  (+ (vector-ref (cdr v) 0)
		     (vector-ref (cdr v) 0))))))))
```

It seems to be working but it definitely needs some more work: we should get rid of the nested `cond`s and remove the redundant tests.

turlando: For a comparison, the generated JS:

```
var Bar = (function () {
    function Bar(value0) {
        this.value0 = value0;
    };
    Bar.create = function (value0) {
        return new Bar(value0);
    };
    return Bar;
})();

var doubleFoo = function (v) {
    return v.value0 + v.value0 | 0;
};
```

turlando: >I am confused about why we throw away identifiers here and in other places.

The reason here is that `bar` is an alias to `v` (thus the comment regarding the level of indirection). `value` is an `Expr.Var "v"`, and ident is an `Ident.Ident "bar"`. We throw the ident away here because we don't need it to generated the test part of this cond clause. But we use it later in `variablesToReplace` in order to make the substitution `bar` -> `v`.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: It is for the same reason we do it at line 171. Should I add a comment nonetheless?

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: It is for the same reason we do it at line 171. Should I add a comment nonetheless?

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: In our case the variable is `v`, but it gets referred as `bar` because of the NamedBinder. Maybe I can make the comment clearer?

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: Summing it up, we have a few things going on here and we should create some cards:
- A function matching a single constructor type should not check for the type, to make the behavior on par with the JS compiler. This is issue #27 but maybe it could be better phrased and described.
- Nested conds should be flattened. Not sure if we should do it by emitting better code or by brutally handling it at the optimizer level.
- Redudant cond clauses should be dropped. This could be tricky.

f-f: > The reason here is that `bar` is an alias to `v` (thus the comment regarding the level of indirection). `value` is an `Expr.Var "v"`, and ident is an `Ident.Ident "bar"`. We throw the ident away here because we don't need it to generated the test part of this cond clause. But we use it later in `variablesToReplace` in order to make the substitution `bar` -> `v`.

Oh I see now, thanks for the clarification - I guess this explanation could go into expanding that comment.

> Summing it up, we have a few things going on here and we should create some cards

I think there should be cards about this, but I'll note that all of this is optimization and thus lower priority than getting all the necessary places in order (e.g. FFI), so I'd say not to worry excessively about this right now

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
f-f: Totally, I missed that occurrence too.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
f-f: I think so, calling this "level of indirection" is a bit confusing without further context - it's important to clarify that it's about the variable names

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: I tried adding a comment but I couldn't think of anything better than "We are handed with an CoreFn Expr. We have to translate in a Scheme expression." Which is pretty obvious.

What does make it feel weird to you?

turlando:
I tried improving the comments. Let me know what you think.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
f-f: Right, I guess my confusion is actually more like "why did we not call `exprToScheme` before and now we do"

f-f (approved):
Looking good!
Let's merge this.

PR #32, opened 2021-06-02
Run tests off a real spago project

### LINKED ISSUE #21: Run tests off a real spago project

turlando:
Fixes #21


src/TestUtil.hs
```diff
@@ -0,0 +1,93 @@
...
+removeCorefnFiles :: IO ()
+removeCorefnFiles = catch (Turtle.rmtree corefnPath) handler
+  where
+    handler :: IOException -> IO ()
+    handler _e = pure () -- No such file or directory
+
+removeSchemeFiles :: IO ()
+removeSchemeFiles = do
+  xs <- Turtle.fold (Turtle.find (Turtle.ends ".sls") schemePath) Fold.list
+  mapM_ Turtle.rm xs
```
f-f, line 88:
```suggestion
```

As we discussed in the call, let's not remove files here in the executable. This is because it might lead to unexpected deletions - the `Makefile` is the best place to put this if we want such a thing

f-f (changes requested):
Looking great! Just one change as we discussed in the last call

src/TestUtil.hs
turlando: I trust make and shell string/path expansion much less than I trust Haskell. Let's have it properly documented and leave the user do the nasty job instead.

turlando:
I applied the changes we discussed. I have another question: should we move `TestUtil.hs` from the root to something under the `Language.PureScript.Scheme` namespace? Could that cause some namespace clash?

f-f:
Sounds good!

f-f (approved):

PR #33, opened 2021-07-05
Change AST To S-Expression

turlando:
Currently we're using AST for stuff that's clearly not syntax: Cond, Lambda, Define.

Going a step further, Scheme has no syntax other than primitive literals (symbols, numbers, strings) and lists. That's exactly what we call S-Expressions. (One could say S-Expressions also include dotted pairs `'(x . y)`, but since that's can be achieved with just `(cons 'x 'y)` we're not including them.)

What we lose: barely nothing. Using Cond to express `cond` expressions gives us no advantage, and the new design can handle conds with a final else expression in a clearer way. Using Lambda to express an anonymous function doesn't allow us to express functions with multiple parameters, which are not a thing in PureScript but they're definitely present in Scheme.

Overall it can give us more flexibility and the code change is rather small (+ 142 / - 167).

It might make the pretty printer slightly complex to make if we want to discriminate over AST, but it would just be an illusion and would bring us to add non-AST stuff to the AST.

### DIFF
--- a/src/Language/PureScript/Scheme/CodeGen/Optimizer.hs
+++ b/src/Language/PureScript/Scheme/CodeGen/Optimizer.hs
@@ -4,30 +4,25 @@ import Language.PureScript.Scheme.CodeGen.SExpr (SExpr(..), everywhere)
 import Language.PureScript.Scheme.CodeGen.Scheme (app)
 
 
-optimizations :: SExpr -> SExpr
-optimizations = (simplifyLogic . specializeOperators)
-
-
 -- Single pass optimizer.
 -- Run through all the SExpr expressions and apply the optimizations.
 runOptimizations :: [SExpr] -> [SExpr]
 runOptimizations xs = map (\x -> everywhere optimizations x) xs
 
 
--- TODO: possibly nasty hack. To review once externs are implemented.
-specializeOperators :: SExpr -> SExpr
+optimizations :: SExpr -> SExpr
+optimizations = (simplifyLogic . inlineCommonBinaryOperators)
 
-specializeOperators (List [List [List [Symbol "Data.Semiring.add",
-                                       Symbol "Data.Semiring.semiringInt"],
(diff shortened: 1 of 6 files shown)


src/Language/PureScript/Scheme/CodeGen/Optimizer.hs
```diff
@@ -1,28 +1,31 @@
...
-specializeOperators :: AST -> AST
+specializeOperators :: SExpr -> SExpr
 
-specializeOperators (Application (Application (Application (Identifier "Data.Semiring.add")
-                                 [Identifier "Data.Semiring.semiringInt"]) [x]) [y])
-  = Application (Identifier "+") [x, y]
+specializeOperators (List [List [List [Symbol "Data.Semiring.add",
+                                       Symbol "Data.Semiring.semiringInt"],
+                           x], y])
+  = app "+" [x, y]
```
f-f, line 24: My experience when working with strongly typed languages is that "more intermediate representations is always better".
This is because "reusing types" is dangerous in a similar way to "reusing code" in that it risks being a poor fit for all the usecases involved where this doesn't happen with specialized code/types.
Another issue with having too few representations is that one risks having abrupt conversions, making the implementation harder to debug.

In this case we might be hitting both issues: doing optimizations on a `AST` that has `Application` is definitely less error prone than working on something that has just a `List` of stuff, and even if this doesn't seem clear cut here, future-you might agree.

So instead of replacing the previous `AST` type with `S-Expr` we might want to have both instead!
In this way:
- we could define optimizations on the more well-typed `AST`
- have a straightforward conversion from `AST` to `S-Expr`
- define the pretty printing only on `S-Expr`

f-f (approved):
The only comment I left is about how I would do this myself.
It should be very easy to address if you'd like, but it's not critical that you address or change the implementation if you don't want to, so I'll approve this and you can just merge if you'd like.

PR #37, opened 2021-07-14
Handle ObjectLiteral

### LINKED ISSUE #36: Handle ObjectLiteral

turlando:
Depends on #35

Fix #36

### DIFF
--- a/src/Language/PureScript/Scheme/CodeGen/Scheme.hs
+++ b/src/Language/PureScript/Scheme/CodeGen/Scheme.hs
@@ -18,6 +18,12 @@ t = Symbol "#t"
 f :: SExpr
 f = Symbol "#f"
 
+stringEqQ' :: SExpr
+stringEqQ' = Symbol "string=?"
+
+stringHash' :: SExpr
+stringHash' = Symbol "string-hash"
+
 
 -- Scheme special forms --------------------------------------------------------
 
--- a/src/Language/PureScript/Scheme/CodeGen/Scheme.hs
+++ b/src/Language/PureScript/Scheme/CodeGen/Scheme.hs
@@ -110,6 +110,16 @@ vector xs = app "vector" xs
 vectorRef :: SExpr -> SExpr -> SExpr
 vectorRef v i = app "vector-ref" [v, i]
 
+-- | (make-hashtable hash equiv?)
+-- hash and equiv? must be procedures. If size is provided, it must be a
+-- nonnegative exact integer indicating approximately how many elements
+-- the hashtable should initially hold.
+makeHashtable :: SExpr -> SExpr -> SExpr -> SExpr
+makeHashtable hash equivQ size = app "make-hashtable" [hash, equivQ, size]
+
+hashtableSetB :: SExpr -> SExpr -> SExpr -> SExpr
+hashtableSetB hashtable key obj = app "hashtable-set!" [hashtable, key, obj]
+
 errorWithWhoAndIrritants :: SExpr -> SExpr -> [SExpr] -> SExpr
 errorWithWhoAndIrritants who msg irritants
   = app "error" $ [who, msg] <> irritants
--- a/src/Language/PureScript/Scheme/CodeGen/Scheme.hs
+++ b/src/Language/PureScript/Scheme/CodeGen/Scheme.hs
@@ -46,6 +46,9 @@ let_ bodies expr = List [Symbol "let", List $ fmap go bodies, expr]
     go :: (Text, SExpr) -> SExpr
     go (var, expr') = List [Symbol var, expr']
 
+let1 :: (Text, SExpr) -> SExpr -> SExpr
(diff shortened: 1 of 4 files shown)


turlando:
The code calls them ObjectLiteral, the documentation calls them [records](https://github.com/purescript/documentation/blob/master/language/Records.md). Should the test go in Test.Literal.ObjectLiteral or should we keep it separated since it involves more related stuff (accessor, updater, ...)?

turlando:
String printing has to be fixed too. Should we do it in this PR or make another one?

f-f:
@turlando 

> String printing has to be fixed too. Should we do it in this PR or make another one?

Should be done in another PR, and we should merge that first

> Should the test go in Test.Literal.ObjectLiteral or should we keep it separated since it involves more related stuff (accessor, updater, ...)?

No preference, but smaller tests are usually better

turlando:
>>String printing has to be fixed too. Should we do it in this PR or make another one?

>Should be done in another PR, and we should merge that first

I created another issue just for that #38

turlando:
Rebased off #39

src/Language/PureScript/Scheme/CodeGen/Scheme.hs
```diff
@@ -15,19 +15,45 @@ app name args = List ((Symbol name):args)
...
+  where
+    go :: (Text, SExpr) -> SExpr
+    go (var, expr') = List [Symbol var, expr']
+
+-- (cond (test expr) ... (else expr))
+condWithElse :: [(SExpr, SExpr)] -> SExpr -> SExpr
+condWithElse clauses elseExpr = cond (clauses ++ [(Symbol "else", elseExpr)])
+
+-- (cond (test expr) ...)
+cond :: [(SExpr, SExpr)] -> SExpr
```
f-f, line 51: Is there any value in having a `cond` without an `else` clause?

If our implementation is wrong and we don't match on any of the clauses then this will presumably return `nil`, which sure is going to be _hard_ to debug?

I'd prefer if we'd always emit an `else` clause, where we'd throw an "internal error" when we don't have an `else` clause at the application level. In this way if our implementation is wrong we'd have a pointer at where things blow up.

In other words, I think `cond` should be implemented with `condWithElse`, where we pass an `else` case that throws.

f-f (approved):
One small comment, but otherwise great stuff!

src/Language/PureScript/Scheme/CodeGen/Scheme.hs
turlando: I agree with your suggestion to always use `condWithElse` instead of `cond`.

As of now we're caling `cond` (without else clause) only at [`Transpiler.hs:129`](https://github.com/turlando/purescm/blob/1f7f27f70da735134de1804b8ed9802610e1a45b/src/Language/PureScript/Scheme/CodeGen/Transpiler.hs#L129) when handling case expressions.

In this case I think we're safe to have an exception raised if we hit the else branch.

I would still keep `cond` in `Scheme.hs` though since the cond (without else) is effectively possible in Scheme and provides a tiny lower abstraction for `condWithElse`. Maybe we can make it private?

turlando:
Some things I'm noticing looking the code again:

- binderToTest can't handle string literals (I missed it in the previous PR)
- Maybe write `let1` for uses such as `Transpiler.hs:115`.
- Maybe factor out hashmap creation in a separate function?

src/Language/PureScript/Scheme/CodeGen/Scheme.hs
f-f: > Maybe we can make it private?

I don't see the point in keeping an unsafe function even if private. What I mean by "I think `cond` should be implemented with `condWithElse`" (instead of the other way around as it is now) is this:

```haskell
condWithElse :: [(SExpr, SExpr)] -> Maybe SExpr -> SExpr
condWithElse clauses maybeElseExpr
  = app "cond"
  $ map (\(test, expr) -> List [test, expr])
  $ clauses <> case maybeElseExpr of
      Nothing -> [Symbol "raise", String "Internal error while handling expression etc etc"]
      Just elseExpr -> [Symbol "else", elseExpr]

cond :: [(SExpr, SExpr)] -> SExpr
cond clauses = condWithElse clauses Nothing
```

In this way we could export both `cond` and `condWithElse`, and they would both be safe :slightly_smiling_face:

f-f:
> 1. binderToTest can't handle string literals (I missed it in the previous PR)
> 2. Maybe write let1 for uses such as Transpiler.hs:115?
> 3. Maybe factor out hashmap creation in a separate function?

Of all of these I think only (1) is critical and should be fixed ASAP, the others I don't have an opinion on but you can go ahead and fix them here if you'd like

turlando:
The first issue should be fixed with #40.

f-f:
Great!

Then I think we are good to merge this after we take care of:
- rebasing/conflicts
- the `cond` stuff

PR #58, opened 2021-09-08
Add --run option

turlando:


test/resources/scheme/PureScheme.Test.Recursion.Mutual/lib.sls
```diff
@@ -1,32 +1,32 @@
...
-          ((= v 1) ((foo 2) v1))
-          ((= v 2) v1)
-          (#t 0)
-          (else (error #f "Failed pattern match"))))))
+        (scm:cond
+          ((scm:and (scm:= v 0) #t) ((bar 1) v1))
+          ((scm:and (scm:= v 1) #t) ((foo 2) v1))
+          ((scm:and (scm:= v 2) #t) v1)
+          ((scm:and #t #t) 0)
+          (scm:else (scm:error #f "Failed pattern match"))))))
```
f-f, line 31: The implementation of `cond` changed?

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -403,14 +422,14 @@ moduleToLibrary (Module _sourceSpan _comments moduleName _path
...
-              (:) (runIdent ident, value)
+              (:) (identToJs ident, value)
                   (go value binder)
 
     ----------------------------------------------------------------------------
 
     letToScheme :: [Bind Ann] -> Expr Ann -> SExpr
     letToScheme binds body
-      = let_ (concatMap bindToScheme binds) (exprToScheme body)
+      = letRecS (concatMap bindToScheme binds) (exprToScheme body)
```
f-f, line 432: Let's make a ticket to investigate performance optimizations of using `let` vs `letrec` vs `let*` vs `letrec*`

f-f: I see that we're synchronizing, this is now tracked in #60 :slightly_smiling_face:

app/Run.hs
```diff
@@ -0,0 +1,44 @@
...
+parseModuleAndFunction :: Text -> (Text, Text)
+parseModuleAndFunction s =
+  let
+    s' = Text.splitOn "." s
+    moduleName = Text.intercalate "." (init s')
+    functionName = last s'
+  in (moduleName, functionName)
+
+run :: Text -> Text -> IO ()
+run outputPath moduleAndFunction = do
```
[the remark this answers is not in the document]
turlando: `spago run -m FOO` wants a module name (e.g. `Main`), but then it calls the backend as `/tmp/purescm --run Main.main`. So here we should really expect a fully qualified function name, and not a module name.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
turlando: #60 here it goes

test/resources/scheme/PureScheme.Test.Recursion.Mutual/lib.sls
turlando: No, what's happening here is that we're missing the simplification optimization for `and`.
We pattern match on function application of `and`, but now we have to match on `scm:and`.

app/Run.hs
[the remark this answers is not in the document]
f-f: My point is that we should not be parsing the function name out of this, as we always expect only a Module name, the `.main` should only be appended when we call out to Scheme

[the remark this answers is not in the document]
f-f: I also suspect that we do this parsing thing because this supports arbitrary function names, but we could remove the parsing once we hardcode `main`

test/resources/scheme/PureScheme.Test.Recursion.Mutual/lib.sls
f-f: I see. I think we should fix this here so that the generated code doesn't change this much

f-f (approved):

f-f:
Great work! :clap:

PR #62, opened 2021-09-29
Add random identifiers when emitting errors for failed pattern matches

f-f:


turlando:
Looks good! If you're ok with that I will remove the changes to `spago.dhall` to this commit and force-push to this branch.

PR #68, opened 2021-10-25
Unify the two conds generated by case binders and guards

### LINKED ISSUE #63: Broken pattern matching with guards
The following code, taken from [List](https://github.com/purescript/purescript-lists/blob/v6.0.1/src/Data/List.purs#L521-L521)

```
take :: forall a. Int -> List a -> List a
take = go Nil
  where
  go acc n _ | n < 1 = reverse acc
  go acc _ Nil = reverse acc
  go acc n (x : xs) = go (x : acc) (n - 1) xs
```

is translated to

```
(define take
  (letrec*
     ([go
       (lambda (acc)
         (lambda (v)
           (lambda (v1)
             (cond
              [#t (cond
                   [(fx<? v 1) (reverse acc)]
                   [else (error #f "Failed pattern match kvtleflvhe")])]
              [(eq? (car v1) (quote Nil)) (reverse acc)]
              [(eq? (car v1) (quote Cons))
               (((go ((Data.List.Types.Cons (vector-ref (cdr v1) 0)) acc)) (fx- v 1)) (vector-ref (cdr v1) 1))]
              [else (error #f "Failed pattern match wahyysksgx")]))))])
    (go Data.List.Types.Nil)))
```

and it always triggers the error for `kvtleflvhe`.

f-f:
Fix #63 as per title. Since there can be multiple guards for each condition branch of a `case`, we duplicate them when merging the conditions. This can be seen in action in the `PureScheme.Test.Binder.Guard` test.

We include the failing example from #63 as a test here, and it now reads correct:
```scheme
  (scm:define
    take
    (scm:letrec*
      ((go
        (scm:lambda (acc)
          (scm:lambda (v)
            (scm:lambda (v1)
              (scm:cond
                ((scm:fx<? v 1) acc)
                ((scm:eq? (scm:car v1) (scm:quote Nil)) acc)
                ((scm:eq? (scm:car v1) (scm:quote Cons))
                  (((go ((Cons (scm:vector-ref (scm:cdr v1) 0)) acc))
                      (scm:fx- v 1))
                    (scm:vector-ref (scm:cdr v1) 1)))
                (scm:else (scm:error #f "Failed pattern match"))))))))
      (go Nil)))
```

### DIFF
--- a/src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
+++ b/src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
@@ -235,32 +235,34 @@ moduleToLibrary (Module _sourceSpan _comments moduleName _path
     caseToScheme values caseAlternatives =
       cond clauses
       where
-        clauses = map condClause alternatives
+        clauses = concatMap condClause alternatives
         alternatives = toAlternatives values caseAlternatives
 
-        -- Emit a pair (condition, result) for each case alternative to be used
-        -- by cond.
-        condClause :: Alternative -> (SExpr, SExpr)
+        -- Emit a list of pairs (condition, result) for each case alternative to be used
+        -- by cond. Every case alternative can produce multiple of them because there
+        -- might be multiple guards for each of them.
+        condClause :: Alternative -> [(SExpr, SExpr)]
 
-        -- If the result is a guarded expression then the result of the clause
-        -- is a nested cond expression.
+        -- If the result is a guarded expression then we concatenate the guard
+        -- conditions to the conditions for the alternative.
         condClause (Alternative boundValues (Left guardedExprs))
-          = ( test boundValues
-            , cond $ map guardedExprToClause guardedExprs
-            )
+          = map guardedExprToClause guardedExprs
           where
             guardedExprToClause :: GuardedExpr -> (SExpr, SExpr)
             guardedExprToClause (guard, result)
-              = ( replaceVariables (exprToScheme guard)
-                                   (variablesToReplace boundValues)
-                , replaceVariables (exprToScheme result)
-                                   (variablesToReplace boundValues)
+              = ( test boundValues $ Just $
+                  replaceVariables
+                    (exprToScheme guard)
+                    (variablesToReplace boundValues)
+                , replaceVariables
+                    (exprToScheme result)
+                    (variablesToReplace boundValues)
                 )
 
         -- If the result is a simple expression then the result is just the
         -- expression.
         condClause (Alternative boundValues (Right result))
-          = (test boundValues, bindersToResult boundValues result)
+          = [(test boundValues Nothing, bindersToResult boundValues result)]
 
         -- Emit the test for a cond clause.
         -- In scheme: (cond ((and test1a test1b) result1) ...)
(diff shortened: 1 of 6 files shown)


src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -285,11 +287,12 @@ moduleToLibrary (Module _sourceSpan _comments moduleName _path
         --     (cond ((and (eq? (car v) 'Foo) cond1)  result1)
         --           ((and (eq? (car v) 'Foo) cond2)) result2)
         -- TODO: Maybe we should avoid the repeating of the first check.
-        test :: [(Expr Ann, Binder Ann)] -> SExpr
-        test valueAndBinders =
-          and_ (concatMap (\(value, binder) -> binderToTest (exprToScheme value)
-                                                            binder)
-                          valueAndBinders)
+        test :: [(Expr Ann, Binder Ann)] -> Maybe SExpr -> SExpr
+        test valueAndBinders maybeGuard = and_ $
```
turlando, line 291: I think it's better not to overload this function with the extra maybeGuard parameter and have a separate function (for instance testWithGuard) to handle guarded expressions.

Another option would be to just leave this function as it was and just append the guard in guardedExprToClause:

```haskell
            guardedExprToClause :: GuardedExpr -> (SExpr, SExpr)
            guardedExprToClause (guard, result)
              = let tests = test boundVariables
                    guard' = replaceVariables (exprToScheme guard)
                                              (variablesToReplace boundValues)
                    tests = boundValues <> guard'
                    result = replaceVariables (exprToScheme result)
                                              (variablesToReplace boundValues)
                in ( tests <> guard'
                   , result
                   )
```

What do you think? If we want to stick with overloading the test function it's better to update its comment.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
```diff
@@ -235,32 +235,34 @@ moduleToLibrary (Module _sourceSpan _comments moduleName _path
...
-        -- by cond.
-        condClause :: Alternative -> (SExpr, SExpr)
+        -- Emit a list of pairs (condition, result) for each case alternative to be used
+        -- by cond. Every case alternative can produce multiple of them because there
+        -- might be multiple guards for each of them.
+        condClause :: Alternative -> [(SExpr, SExpr)]
 
-        -- If the result is a guarded expression then the result of the clause
-        -- is a nested cond expression.
+        -- If the result is a guarded expression then we concatenate the guard
```
turlando, line 246: Concatenation happens in caseToScheme, here we just generate a list of clauses.

turlando (commented):
Great solution, I just added few notes open to discussion.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
f-f: We cannot just append the guard in `guardedExprToClause`, because `test` returns an opaque `SExpr` (as it wraps everything in `and_`), so whatever we want to put in there we need to pass it in.
I also understand your concern about "overloading", but consider the fact that we're not really expanding the scope of the `test` function: it's still supposed to produce the test for a single cond clause. Since this test is produced slightly differently in the case of guarded expressions I think it makes sense to pass the additional data in. On a similar line, I don't think it's worth splitting this in a `testWithGuard` function, since there's only one switch that amounts to three lines.

All in all, I think that this implementation is the least dangerous patch and we should stick to it - I'll update the comment to document the new behaviour.

src/Language/PureScript/Scheme/CodeGen/Transpiler.hs
f-f: No, the concatenation happens here because produce a `SExpr` for the guard, and we pass it to `test`

PR #69, opened 2023-03-08
Relicense to Apache-2.0

f-f:
Affero scares people away it looks like 🙁 

@turlando we're the only two authors and this will need your explicit approval before merge.


turlando:
> Affero scares people away it looks like

Yeah, it's a feature.

JordanMartinez:
I didn't expect this change to happen so fast. I've read through @f-f's explanation for why it was used initially, and I've just now read through a few other posts on the matter.

I'm ok with contributing towards this project under AGPL-3.0. While others in Slack indicated their initial agreement with me in only using permissive licenses, the question remains whether they will actually contribute anything.

@purefunctor, since you expressed interest in this, would you be willing to contribute under AGPL-3.0?

purefunctor:
> I didn't expect this change to happen so fast. I've read through @f-f's explanation for why it was used initially, and I've just now read through a few other posts on the matter.
> 
> I'm ok with contributing towards this project under AGPL-3.0. While others in Slack indicated their initial agreement with me in only using permissive licenses, the question remains whether they will actually contribute anything.
> 
> @purefunctor, since you expressed interest in this, would you be willing to contribute under AGPL-3.0?

I'm fine with either license, I guess my only concern is compatibility if the project gets upstreamed

f-f:
Thanks for the support folks - but I personally think we should stick with Apache - while AGPL is the most restrictive open source out there (which means "good for open source"), it's also fairly "radioactive" in the sense that a lot of corporate folks will refuse to come close to it (Google is a famous example), and "contagious", in the sense that it would force people to release things under AGPL even if they are communicating with the software over the wire.
I am happy about this as an open source citizen, but much less happy about it as a "language steward", in the sense that such a restrictive license would prevent the spread of the software we write.

Issue #74, opened 2023-03-15
Pretty-print generated Chez Scheme output


purefunctor:
Right now, not much pretty-printing is being done for generated Scheme sources. A possible approach to take is to create a separate syntax for the pretty-printer such that we wouldn't have to wrangle with `Dodo` primitives or pattern matching against the core syntax.

JordanMartinez:
I'll submit a PR later tonight or tomorrow morning that adds some pretty printing by redefining the Chez Scheme syntax to better adhere to the spec.

JordanMartinez:
Should we model [`expr`](https://github.com/JordanMartinez/purs-backend-cs/blob/master/chez-scheme-eBNF.md?plain=1#L119-L127)? If so, how much of it should we model (I'm also thinking how well/poorly this will help when it comes to parsing FFI)?
Or should we continue to use the current `ChezExpr` type?

#76 is the PR I mentioned above.

JordanMartinez:
Small nitpick, but do ya'll mind if we change the alias for `Dodo` to just `D`? I have a habit of writing `D.text`, but here I need to write `Dodo.text`. If so, I can do that in a separate PR.

purefunctor:
> Should we model [`expr`](https://github.com/JordanMartinez/purs-backend-cs/blob/master/chez-scheme-eBNF.md?plain=1#L119-L127)? If so, how much of it should we model (I'm also thinking how well/poorly this will help when it comes to parsing FFI)? Or should we continue to use the current `ChezExpr` type?
> 
> #76 is the PR I mentioned above.

So far I haven't run anything that the `ChezExpr` type isn't able to model (aside from quoted stuff, I guess), but I think it's reasonable to model `expr` for completion.

> Small nitpick, but do ya'll mind if we change the alias for `Dodo` to just `D`?

Do go for it!

JordanMartinez:
Closed by #76 and #79.

> What's the rationale for modeling the whole expr? If this is output-only (i.e. only printing out code) then we don't really need much syntax at all

I think we can further discuss this in a separate issue, but I think you're right: for output-only, we don't need the `expr` syntax.

PR #76, opened 2023-03-16
Refactor syntax towards spec; pretty-print things

JordanMartinez:
Begins to address #74

I'm following the spec I summarized in https://github.com/JordanMartinez/purs-backend-cs/blob/master/chez-scheme-eBNF.md

I haven't fully implemented the types for the `expr`, but this is at least a better start than what we had.

I haven't removed the other code PureFunctor wrote in case we want to reuse parts of that.

### DIFF
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -4,13 +4,13 @@ import Prelude
 
 import Data.Argonaut as Json
 import Data.Array as Array
+import Data.Array.NonEmpty as NEA
 import Data.Array.NonEmpty as NonEmptyArray
 import Data.Maybe (Maybe(..))
-import Data.Newtype (wrap)
-import Data.Profunctor.Strong (second)
+import Data.Newtype (un, unwrap, wrap)
 import Data.String.CodeUnits as CodeUnits
-import Data.Tuple (Tuple, uncurry)
-import PureScript.Backend.Chez.Syntax (ChezExpr)
+import Data.Tuple (Tuple(..), uncurry)
+import PureScript.Backend.Chez.Syntax (ChezDefinition(..), ChezExport(..), ChezExpr, ChezImport(..), ChezImportSet(..), ChezLibrary)
 import PureScript.Backend.Chez.Syntax as S
 import PureScript.Backend.Optimizer.Convert (BackendModule, BackendBindingGroup)
 import PureScript.Backend.Optimizer.CoreFn (Ident(..), Literal(..), ModuleName(..), Qualified(..))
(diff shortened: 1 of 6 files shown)


f-f:
This is a nit and I'm not sure how hard it is to implement, but traditionally in lisp the closing parentheses of a form goes together with the last item, so e.g.
```scheme
(scm:define number
  42.0)
```

rather than

```scheme
(scm:define number
  42.0
)
```

JordanMartinez:
> This is a nit and I'm not sure how hard it is to implement

That's not hard to fix. How's the latest commit?

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -22,28 +24,106 @@ type CodegenEnv =
...
+codegenModule :: BackendModule -> ChezLibrary
 codegenModule { name, bindings, exports } =
   let
     codegenEnv :: CodegenEnv
     codegenEnv = { currentModule: name }
 
-    exports' :: Array Ident
-    exports' = Array.fromFoldable exports
+    exports' :: Array ChezExport
+    exports' = map (ExportIdentifier <<< coerce) $ Array.fromFoldable exports
```
purefunctor, line 34:
```suggestion
    exports' = map ExportIdentifier $ coerce $ Array.fromFoldable exports
```

purefunctor (changes requested):
LGTM!

src/PureScript/Backend/Chez/Convert.purs
JordanMartinez: Done.

f-f (approved):
Looking great, thanks!

test-snapshots/snapshots-output/Snapshot.Literals.ss
```diff
@@ -2,7 +2,7 @@
 #!chezscheme
 (library
   (Snapshot.Literals lib)
-  (exports
+  (export
```
JordanMartinez, line 5: Latest commit: this was a typo.

test-snapshots/snapshots-output/Snapshot.Literals.ss
```diff
@@ -14,8 +14,8 @@
     record2
     string
   )
-  (imports
-    (prefix (library (chezscheme)) scm:)
+  (import
```
JordanMartinez, line 17: Latest commit: this was a typo.

test-snapshots/snapshots-output/Snapshot.Literals.ss
```diff
@@ -14,8 +14,8 @@
...
+    (prefix ((chezscheme)) scm:)
```
JordanMartinez: Latest commit: `imports` was a typo.

Also, after re-reading that part in the R6RS spec, `library` is only needed if the first identifier is one of the keywords that could be confused for an import set. When I copy-pasted this file in the Scheme REPL, it rejected it until I made this change.

JordanMartinez:
Ok. I've double-checked things according to the R6RS report, and I believe the types and code for a library now adhere to the spec.

purefunctor:
I'm trying to run the test script locally but for some reason I'm getting an `unsafePartial` error for when the backend is attempting to compile `Data.Unit`

test/Main.purs
```diff
@@ -108,7 +108,7 @@ runSnapshotTests { accept, filter } = do
               formatted =
                 Dodo.print plainText Dodo.twoSpaces
                   $ flip append Dodo.break
```
purefunctor, line 110: I think we can move this ending newline to the `printLibrary` function now

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -22,28 +24,106 @@ type CodegenEnv =
...
-    exports' :: Array Ident
-    exports' = Array.fromFoldable exports
+    exports' :: Array ChezExport
+    exports' = map ExportIdentifier $ coerce $ Array.fromFoldable exports
 
-    bindings' :: Array ChezExpr
-    bindings' = Array.concat $ codegenTopLevelBindingGroup codegenEnv <$> bindings
+    definitions :: NonEmptyArray ChezDefinition
+    definitions = unsafePartial $ fromJust $ NEA.fromArray $ Array.concat $
+      codegenTopLevelBindingGroup codegenEnv <$> bindings
```
purefunctor, line 38: Given an empty module, this crashes:
```purescript
module Snapshot.Empty where

```

Also caught this on https://github.com/purescm/purescm/pull/76#issuecomment-1473872909 for some reason, but that could just be an oddity with build caching.

There's two ways to go about this:
1. Create a placeholder `define` such that `exprs` can still be inserted
2. Drop the `exprs` field in the `body` and revert to a regular `Array`

PureScript doesn't have expressions at the top-level, so 2. seems the most reasonable for me...

purefunctor (changes requested):

src/PureScript/Backend/Chez/Convert.purs
JordanMartinez: The code above is just wrong. The spec says the `libraryBody` is
`〈definition〉 ... 〈expression〉 ...`, which means both `definition` and `expression` can both be empty.

I think I thought that was weird because why have a library that doesn't have anything inside of it? On a second reading, that also means the `export` and `import` lists can be empty. But again, why have a library if it doesn't export anything?

I guess my understanding is just flawed because libraries can also execute code found in the library body's `exprs`.

JordanMartinez: Fixed.

JordanMartinez:
There are a few other places where a non-empty array is used as the type here, but that's not what's in the R6RS spec. (See Section 7 in http://www.r6rs.org/final/r6rs.pdf). 

In the last paragraph of Section 6.2, they clarify that they use `<foo> ...` to indicate zero or more of `foo` and `<foo1> <foo2> ...` to indicate one or more of `foo`.

Examples of the eBNF grammars in the Library section that confuse me are:
- `import` allows two possibilities: `<import-set>` and `(for <import-set> <import-level>...)`. But the second means one could write `(for <import-set>)` without the level. But why would someone do that if it's the same as just writing `<import-set>`? So, I opted to make the array non-empty, but should it be that way?
- `import-set` allows `<import set>`, `(only 〈import set〉 〈identifier〉 ...)`, and other such options. Again, I'm confused about this for the same reasons as above.

src/PureScript/Backend/Chez/Convert.purs
JordanMartinez: > There's two ways to go about this:
> 1. Create a placeholder define such that exprs can still be inserted
> 2. Drop the exprs field in the body and revert to a regular Array

I think we should keep `exprs` because we can use the same type when parsing FFI files.

purefunctor: Would it be appropriate to encode the invariant that `exprs` can only be non-empty if `definitions` is non-empty as well?

JordanMartinez: This is weird... Running the following in the REPL works:
```scheme
(library (a) (export) (import))
```

JordanMartinez:
Well... for the mean time, I updated the types to allow for empty arrays. It adheres to the spec, but it makes no sense to me. :shrug:

JordanMartinez:
Last commit renames the alias from `Dodo` to `D` and `D.Doc` to just `Doc`.

JordanMartinez:
I don't intend to make any more changes on this PR.

purefunctor (approved):

JordanMartinez:
Oh, one other question. Do you guys have a preference for how we do merges in this repo? In most other PS repos, we do squash merges.

JordanMartinez:
Github:
> This branch cannot be rebased due to conflicts

Let me fix that.

Issue #71, opened 2023-03-15
Add CI


f-f:
The existing workflows are Haskell based, but we need build PureScript now.

purefunctor:
Finished since #109

PR #78, opened 2023-03-18
Install CS in CI, following Idris2's approach

JordanMartinez:
Second part of #71. This follows the approach taken by Idris2:
- [Ubuntu](https://github.com/idris-lang/Idris2/blob/main/.github/workflows/ci-idris2.yml#L135-L161)

Snapshot files are now evaluated via `scheme --script path-to-output.ss`. If there are any issues with the codgen'd output, CI will now complain.

When we have more working FFI, we can implement run the snapshots and verify that they work and produce the expected output.


JordanMartinez:
One issue that I'm noticing is that the Chez Scheme versions installed by each OS is slightly different:
- Linux: 9.5.4
- Mac 9.5.8
- Windows: 9.5.9

.github/workflows/ci.yml
```diff
@@ -7,43 +7,96 @@ on:
...
-            output
-
-      - name: Set up Node toolchain
-        uses: actions/setup-node@v2
-        with:
-          node-version: "16"
+      - name: Install chez scheme
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y chezscheme
```
JordanMartinez, line 18: One issue that I'm noticing is that the Chez Scheme versions installed by each OS is slightly different:
- Linux: 9.5.4
- Mac 9.5.8
- Windows: 9.5.9

I assume we want to standardize on a single version across CI. Thoughts?

.github/workflows/ci.yml
```diff
@@ -7,43 +7,96 @@ on:
...
+      
+      - name: Verify Chez Scheme installed
+        run: |
+          ${SCHEME} --version
```
JordanMartinez, line 22: Any problems this approach for running the binary correctly?
- Linux: scheme
- Mac: chez
- Windows: scheme

.github/workflows/ci.yml
```diff
@@ -7,43 +7,96 @@ on:
...
+        run: |
+          c:\msys64\usr\bin\bash -l -c "cd $env:PWD && cd ChezScheme && ./configure --threads && make"
+      - name: Set Path
         run: |
-          purs-tidy check src test
+          $chez="$(pwd)\ChezScheme\ta6nt\bin\ta6nt"
+          echo "$chez" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+      - name: Test Scheme
+        run: |
+          Invoke-Expression "$env:SCHEME --version"
```
JordanMartinez, line 62: I don't know if I like this approach because I'd guess any future appearance of double quotes (e.g. for a CLI arg) will be brittle. However, due to my unfamiliarity with Windows, I'm not sure how else to invoke the env var.

JordanMartinez:
The next few commits will focus on updating the snapshots to use the binary.

packages.dhall
```diff
@@ -70,4 +70,5 @@ let additions =
         }
       }
 
-in  upstream // additions
+in  (upstream // additions)
+      with node-execa.version = "v1.2.0"
```
JordanMartinez, line 74: While testing this locally, I found a bug in `node-execa`'s FFI because I forgot to [add the `return` keyword](https://github.com/JordanMartinez/purescript-node-execa/commit/bf5afc51a8111f8f922fa6a09b3f105802c865d9). This is a situation where being able to parse the JS FFI and warn me that my function returns nothing would have been helpful.

JordanMartinez:
Technically, we don't need the 3 OSes until after we're ready to compile this tool to CS itself. So, this PR just uses Ubuntu to build, test, and lint our code.

f-f (approved):

PR #81, opened 2023-03-20
Implement Branch, PrimOp, and Fail; add runtime module

purefunctor:
This PR implements code generation for `Branch`, `PrimOp`, and `Fail` (partially, see comment in source). This also adds a runtime module which provides functions that are otherwise not present in the standard library.

### DIFF
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -151,6 +151,9 @@ codegenExpr codegenEnv@{ currentModule } (NeutralExpr s) = case s of
     S.Identifier "prim-undefined"
 
   Fail i ->
+    -- Note: This can be improved by using `error`, but it requires
+    -- that we track where exactly this `Fail` is defined. We can
+    -- make use of the `codegenEnv` for this.
     S.List
       [ S.Identifier "scm:raise"
       , S.List
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -150,8 +150,18 @@ codegenExpr codegenEnv@{ currentModule } (NeutralExpr s) = case s of
   PrimUndefined ->
     S.Identifier "prim-undefined"
 
-  Fail _ ->
-    S.Identifier "fail"
+  Fail i ->
+    S.List
+      [ S.Identifier "scm:raise"
+      , S.List
+          [ S.Identifier "scm:condition"
+          , S.List [ S.Identifier "scm:make-error" ]
+          , S.List
+              [ S.Identifier "scm:make-message-condition"
+              , S.String $ Json.stringify $ Json.fromString i
+              ]
+          ]
+      ]
 
 codegenLiteral :: CodegenEnv -> Literal NeutralExpr -> ChezExpr
 codegenLiteral codegenEnv = case _ of
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -182,8 +182,54 @@ codegenPrimOp codegenEnv = case _ of
         OpGte -> S.Identifier "scm:>="
         OpLt -> S.Identifier "scm:<"
         OpLte -> S.Identifier "scm:<="
+
+      opCharOrd = case _ of
+        OpEq -> S.Identifier "scm:char=?"
+        OpNotEq -> S.Identifier "scm:char=?"
+        OpGt -> S.Identifier "scm:char>?"
+        OpGte -> S.Identifier "scm:char>=?"
+        OpLt -> S.Identifier "scm:char<?"
+        OpLte -> S.Identifier "scm:char<=?"
+
+      opStringOrd = case _ of
+        OpEq -> S.Identifier "scm:string=?"
+        OpNotEq -> S.Identifier "scm:string=?"
+        OpGt -> S.Identifier "scm:string>?"
+        OpGte -> S.Identifier "scm:string>=?"
+        OpLt -> S.Identifier "scm:string<?"
+        OpLte -> S.Identifier "scm:string<=?"
     in
       case o of
+        OpArrayIndex ->
+          S.List [ S.Identifier "scm:vector-ref", x', y' ]
+        OpBooleanAnd ->
+          S.List [ S.Identifier "scm:and", x', y' ]
+        OpBooleanOr ->
+          S.List [ S.Identifier "scm:or", x', y' ]
+        OpBooleanOrd _ ->
+          -- Neither R6RS nor Chez defines comparison operators for
+          -- booleans, meaning that we would either have to inject
+          -- the definitions at the call site _or_ provide them
+          -- through a library loaded by _all_ files.
+          S.Identifier "boolean-ord"
+        OpCharOrd o' ->
+          case o' of
+            OpNotEq ->
+              S.List [ S.Identifier "scm:not", S.List [ opCharOrd o', x', y' ] ]
+            _ ->
+              S.List [ opCharOrd o', x', y' ]
+        OpIntBitAnd ->
+          S.List [ S.Identifier "scm:bitwise-and", x', y' ]
+        OpIntBitOr ->
+          S.List  [ S.Identifier "scm:bitwise-or", x', y' ]
+        OpIntBitShiftLeft ->
+          S.List [ S.Identifier "scm:bitwise-arithmetic-shift-left", x', y' ]
+        OpIntBitShiftRight ->
+          S.List [ S.Identifier "scm:bitwise-arithmetic-shift-right", x', y' ]
+        OpIntBitXor ->
+          S.List [ S.Identifier "scm:bitwise-xor", x', y' ]
+        OpIntBitZeroFillShiftRight ->
+          S.Identifier "zero-fill-shift-right"
         OpIntNum o' ->
           S.List [ opDtNum o', x', y' ]
         OpIntOrd o' ->
@@ -200,5 +246,13 @@ codegenPrimOp codegenEnv = case _ of
               S.List [ S.Identifier "scm:not", S.List [ opDtOrd o', x', y' ] ]
             _ ->
               S.List [ opDtOrd o', x', y' ]
-        _ ->
-          S.Identifier "binary-prim-op"
+        OpStringAppend ->
+          S.List [ S.Identifier "scm:string-append", x', y' ]
+        OpStringOrd o' ->
+          -- We've done this pattern-matching four times. Should this
+          -- be refactored to have the pretty-printer deal with it?
+          case o' of
+            OpNotEq ->
+              S.List [ S.Identifier "scm:not", S.List [ opStringOrd o', x', y' ] ]
+            _ -> 
+              S.List [ opStringOrd o', x', y' ]
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -129,12 +129,11 @@ codegenExpr codegenEnv@{ currentModule } (NeutralExpr s) = case s of
     S.Identifier "let-rec"
   Let i l v e ->
     S.chezLet (S.toChezIdent i l) (codegenExpr codegenEnv v) (codegenExpr codegenEnv e)
-  Branch b o ->
+  Branch b o -> do
     let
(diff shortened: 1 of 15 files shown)


src/PureScript/Backend/Chez/Syntax.purs
```diff
@@ -76,13 +76,13 @@ data ChezImportSet
...
   { definitions :: Prim.Array ChezDefinition
-  , exprs :: Prim.Array ChezExpr
+  , expressions :: Prim.Array ChezExpr
   }
 
 data ChezDefinition
   = DefineValue Prim.String ChezExpr
   | DefineCurriedFunction Prim.String (NonEmptyArray Prim.String) ChezExpr
-  | DefineUncurriedFunction Prim.String (NonEmptyArray Prim.String) ChezExpr
+  | DefineUncurriedFunction Prim.String (Prim.Array Prim.String) ChezExpr
```
purefunctor, line 85: Uncurried functions can have zero arguments, as implied by the `Fn0` type: https://pursuit.purescript.org/packages/purescript-functions/6.0.0/docs/Data.Function.Uncurried#t:Fn0

In Chez, they'd look like this:
```cs
> (define fn0 (lambda () 0))
> (fn0)
0
```

test-snapshots/snapshots-output/Snapshot.Branch.ss
```diff
@@ -0,0 +1,19 @@
...
+  (Snapshot.Branch lib)
+  (export
+    f
+    g)
+  (import
+    (prefix (chezscheme) scm:))
+
+  (scm:define g
+    (scm:lambda (v0)
+      (scm:cond ((prim-op 1) (prim-op 2) (prim-op 3)) (scm:else 0))))
```
purefunctor, line 13: `prim-op` is up next for me

purefunctor:
I think (further) development of the `prelude` port should happen alongside this PR, such that we can test optimizations specific to the core of the language. Of course, I can always cheat and define stubs for modules that have yet to be ported.

I can also imagine that porting would be accelerated if we generated stubs for the foreign files based on the foreign imports in a PureScript module.

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -210,3 +159,100 @@ codegenLiteral codegenEnv = case _ of
...
+        OpMultiply -> S.Identifier "scm:*"
+        OpDivide -> S.Identifier "scm:/"
+
+      opDtOrd = case _ of
+        OpEq -> S.Identifier "scm:="
+        OpNotEq -> S.Identifier "scm:="
+        OpGt -> S.Identifier "scm:>"
+        OpGte -> S.Identifier "scm:>="
+        OpLt -> S.Identifier "scm:<"
+        OpLte -> S.Identifier "scm:<="
```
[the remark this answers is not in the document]
JordanMartinez: Distinguishing between them would probably be better.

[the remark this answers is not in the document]
purefunctor: The code generator now uses `fx`/`fl` functions.

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -210,3 +159,100 @@ codegenLiteral codegenEnv = case _ of
...
+        OpIntBitOr ->
+          S.List  [ S.Identifier "scm:bitwise-or", x', y' ]
+        OpIntBitShiftLeft ->
+          S.List [ S.Identifier "scm:bitwise-arithmetic-shift-left", x', y' ]
+        OpIntBitShiftRight ->
+          S.List [ S.Identifier "scm:bitwise-arithmetic-shift-right", x', y' ]
+        OpIntBitXor ->
+          S.List [ S.Identifier "scm:bitwise-xor", x', y' ]
+        OpIntBitZeroFillShiftRight ->
+          S.Identifier "zero-fill-shift-right"
```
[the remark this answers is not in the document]
purefunctor: I switched this to use `fx` instead.

purefunctor:
The latest iteration of this PR adds a runtime module that includes polyfills for missing functions that we need (e.g. boolean comparison operators)

JordanMartinez:
> The latest iteration of this PR adds a runtime module that includes polyfills for missing functions that we need (e.g. boolean comparison operators)

Can [this line](https://github.com/purescm/purescm/pull/81/commits/eefcdb3b563883ded0bda84aa0cb35222733ec51#diff-71732b478b4808898d86c8591ad7ab46d8122c1e4facec4a9151ac49efba905dR111) be removed then?

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -63,9 +63,12 @@ codegenModule { name, bindings, exports, imports, foreign: foreign_ } =
         [ ImportSet $ ImportPrefix
             (ImportLibrary { identifiers: NEA.singleton "chezscheme", version: Nothing })
             "scm:"
+        , ImportSet $ ImportPrefix
+            (ImportLibrary { identifiers: NEA.cons' "_Chez_Runtime" [ "lib" ], version: Nothing })
```
JordanMartinez, line 67: Warning: Bikeshedding ahead!

Since the library name has to match the file to which the identifier corresponds, is `_Chez_Runtime` what we want to call this? Or should it be named something more specific to PureScript? For example, 
`identifiers: NEA.cons' "purs" [ "backend", "chez-scheme", "lib" ]`

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -191,15 +143,28 @@ codegenExpr codegenEnv@{ currentModule } (NeutralExpr s) = case s of
...
+    -- that we track where exactly this `Fail` is defined. We can
+    -- make use of the `codegenEnv` for this.
+    S.List
+      [ S.Identifier "scm:raise"
+      , S.List
+          [ S.Identifier "scm:condition"
+          , S.List [ S.Identifier "scm:make-error" ]
+          , S.List
+              [ S.Identifier "scm:make-message-condition"
+              , S.String $ Json.stringify $ Json.fromString i
```
JordanMartinez, line 164: Searching for `Fail` (https://github.com/search?q=repo%3Aaristanetworks%2Fpurescript-backend-optimizer+Fail&type=code), Nate may have anticipated a future need where a more descriptive error message was desired per the `unsafeCrashWith "Unsupported fail."` case in the 7th file from the top of that search.

test-snapshots/snapshots-input/Snapshot.Branch.purs
```diff
@@ -0,0 +1,23 @@
...
+g = case _ of
+  0 -> 1
+  1 -> 2
+  2 -> 3
+  _ -> 0
+
+h :: Number -> Number
+h = case _ of
+  3.14 -> 3.14159
+  _ -> 0.0
```
JordanMartinez, line 16: Since you now convert `Fail`, would using a partial pattern match work so we can verify the `Fail` is being outputted correctly?
```
fail :: Partial => Int -> Int
fail = unsafePartial case _ of
  1 -> 1
```

If not, perhaps we shouldn't convert it until ADTs are implemented.

src/PureScript/Backend/Chez/Convert.purs
natefaubion: At one point I was, but I think this is only used for pattern match failure. I will probably remove the string the in the future.

src/PureScript/Backend/Chez/Convert.purs
purefunctor: That would then require the file to be placed under:
```
output/purs/backend/chez-scheme/lib.ss
```
Right?

src/PureScript/Backend/Chez/Runtime.purs
```diff
@@ -0,0 +1,53 @@
...
+          [ DefineUncurriedFunction "boolean->integer" [ "x" ] $ S.List
+              [ S.Identifier "scm:if"
+              , S.Identifier "x"
+              , S.Integer $ S.LiteralDigit "1"
+              , S.Integer $ S.LiteralDigit "0"
+              ]
+          , makeBooleanComparison ">?"
+          , makeBooleanComparison ">=?"
+          , makeBooleanComparison "<=?"
+          , makeBooleanComparison "<?"
```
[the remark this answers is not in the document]
purefunctor: It would actually need to be something like this instead:
```
tell $ [ ExportIdentifier functionName ]
```

I think `StateT` is more apt here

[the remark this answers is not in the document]
JordanMartinez: > It would actually need to be something like this instead:

Ah... right.

> I think StateT is more apt here

Why State? That allows you to modify things whereas Writer does not. Or are you anticipating doing more than just `tell` here in the future?

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -200,57 +200,31 @@ codegenPrimOp codegenEnv = case _ of
...
+          comparisonExpression opName =
+            S.List [ S.Identifier $ Array.fold [ "scm:", tyName, opName ], x', y' ]
+        in
+          case _ of
+            OpEq -> comparisonExpression "=?"
+            OpNotEq -> S.List [ S.Identifier "scm:not", comparisonExpression "=?" ]
+            OpGt -> comparisonExpression ">?"
+            OpGte -> comparisonExpression ">=?"
+            OpLt -> comparisonExpression "<?"
+            OpLte -> comparisonExpression "<=?"
```
JordanMartinez, line 227: Thoughts on using "do...let" here rather than "let...in"?

src/PureScript/Backend/Chez/Convert.purs
JordanMartinez: I believe so.

JordanMartinez: Per Chez Scheme User Guide 9 (Section 2.4):
> For example, for base `/usr/lib/scheme`, library name `(app lib1)`, and extension `.sls`, the full path is `/usr/lib/scheme/app/lib1.sls`. So, if `(library-directories)` contains the pathnames `"/usr/lib/scheme/libraries"` and `"."`, and `(library-extensions)` contains the extensions `.ss` and `.sls`, the path of the `(tools sorting)` library must be one of the following.
> ```
> /usr/lib/scheme/libraries/tools/sorting.ss
> /usr/lib/scheme/libraries/tools/sorting.sls
> ./tools/sorting.ss
> ./tools/sorting.sls 
> ```

src/PureScript/Backend/Chez/Runtime.purs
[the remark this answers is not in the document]
purefunctor: I guess it's just less icky to do `modify_ $ Array.cons ...` than `tell $ [ ... ]` for me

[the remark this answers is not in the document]
JordanMartinez: Well... Icky or not, I think my point still stands.

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -213,18 +211,18 @@ codegenPrimOp codegenEnv = case _ of
...
-            OpLt -> comparisonExpression "<?"
-            OpLte -> comparisonExpression "<=?"
+        case _ of
+          OpEq -> comparisonExpression "=?"
+          OpNotEq -> S.List [ S.Identifier "scm:not", comparisonExpression "=?" ]
+          OpGt -> comparisonExpression ">?"
+          OpGte -> comparisonExpression ">=?"
+          OpLt -> comparisonExpression "<?"
+          OpLte -> comparisonExpression "<=?"
     in
```
JordanMartinez, line 226: Another place where 'do...let' can be used?

JordanMartinez (approved):
I this PR is good enough as is and merging it now unblocks some other things. The following things can be debated/done in a future PR:
- [ ] what the runtime file's name should be
- [ ] whether to use "do...let" or "let...in"
- [ ] removing the [copy foreign file line](https://github.com/purescm/purescm/pull/81/commits/eefcdb3b563883ded0bda84aa0cb35222733ec51#diff-71732b478b4808898d86c8591ad7ab46d8122c1e4facec4a9151ac49efba905dR111)
- [ ] using `D.words`
- [ ] whether to use `WriterT` or `StateT` for exports.

src/PureScript/Backend/Chez/Convert.purs
purefunctor: I personally think the `let` block here is too large for do-let. The `in` is a good separator for when you can't see the `let` and vice-versa.

PR #83, opened 2023-03-24
Add shell.nix mirroring the CI configuration

turlando:
The only difference is that purs-tidy 0.9.0 is installed instead of 0.9.2 due to easy-purescript-nix not providing the newer version yet.


JordanMartinez:
> The only difference is that purs-tidy 0.9.0 is installed instead of 0.9.2 due to easy-purescript-nix not providing the newer version yet.

If there isn't already one, why not submit a PR updating that to `0.9.2` first then?

turlando:
I submitted [easy-purescript-nix/pull/219](https://github.com/justinwoo/easy-purescript-nix/pull/219).

turlando:
Now we should be on par with the CI configuration.

purefunctor (approved):

Issue #84, opened 2023-03-24
Use records instead of hash tables when possible


turlando:
Right now we're translating PS records into Scheme hash table. R6RS and Chez provide records which are supposedly a better fit.

1. [R6RS - Chapter 6: Records](http://www.r6rs.org/final/html/r6rs-lib/r6rs-lib-Z-H-7.html#node_chap_6)
2. [The Scheme Programming Language 4th Edition - Chapter 9: Records](https://www.scheme.com/tspl4/records.html)
3. [Chez Scheme Version 9 User's Guide - Section 7.13: Record Types](https://cisco.github.io/ChezScheme/csug9.5/objects.html#./objects:h13)

purefunctor:
One thing I'm curious about with records for record types is row polymorphism, for example:
```hs
xOrMore :: forall r. { x :: Int | r } -> Int
xOrMore { x } = x

example = xOrMore { x: 0, y: 1 }

example' = xOrMore { x: 0, z: 1 }
```
What would be the generation for this be like?

turlando:
I think we have three options, depending on the information we can get from backend-optimizer:

1. Fall back to hashtables
2. Nest records (I'm kinda against this)
3. Emit a new record for each normalized (?) type (not sure if this is a good idea either)

JordanMartinez:
This problem might need to be solved by adding types to `CoreFn`... A related issue is:
```purs
foo :: forall r. { a :: Int | r } ->  { a :: Int | r }
foo x = x { a = 1 }
```
In the above situation, we need to create a copy of the record with the included update. However, I'm not sure whether `CoreFn` includes all the keys of the record.

natefaubion:
You would have to construct your own inference pass, as @JordanMartinez notes there is no type information in CoreFn or the optimizer codebase. An inference pass would like entail most record arguments being inferred polymorphic. 

I think it's fairly safe to say (for better or worse; for portability worse) that with the current FFI surface area, it's expected that there is at least the ability to reflect record fields to strings and back, which means that records will have to at least contain enough metadata to operate like a hashtable even if there are more efficient access patterns for concrete field selectors.

I would wager that chez scheme will compute hashes for symbols only once, at least? I could be wrong, but I would expect that to be an optimization since symbols are generally interned.

anttih:
We have since changed records to be assoc lists with symbols as keys. This representation is much faster than hash-tables, at least.

PR #85, opened 2023-03-24
Implement data constructors

turlando:
This PR implements code generation for CtorDef and CtorSaturated.

### DIFF
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -138,8 +138,8 @@ codegenExpr codegenEnv@{ currentModule } (NeutralExpr s) = case s of
     S.Identifier $ "accessor-get-prop"
   Accessor _ (GetIndex _) ->
     S.Identifier $ "accessor-get-index"
-  Accessor e (GetOffset o) ->
-    S.recordAccessor (codegenExpr codegenEnv e) o
+  Accessor e (GetCtorField qi _ _ _ field _) ->
+    S.recordAccessor (codegenExpr codegenEnv e) (S.resolve currentModule qi) field
   Update _ _ ->
     S.Identifier "object-update"
 
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -36,6 +36,7 @@ codegenModule { name, bindings, exports: _exports, imports, foreign: foreign_ } 
 
     exports' :: Array ChezExport
     exports' = map ExportIdentifier
+      $ Array.sort
       $ Array.concatMap S.definitionIdentifiers definitions
           <> map coerce (Array.fromFoldable foreign_)
 
@@ -138,7 +139,7 @@ codegenExpr codegenEnv@{ currentModule } (NeutralExpr s) = case s of
   CtorSaturated _ _ _ _ _ ->
     S.Identifier "ctor-saturated"
   CtorDef _ _ _ _ ->
-    unsafeCrashWith "Unreacheable"
+    unsafeCrashWith "codegenExpr:CtorDef - handled by codegenTopLevelBinding!"
 
   LetRec _ _ _ ->
     S.Identifier "let-rec"
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -15,7 +15,7 @@ import Partial.Unsafe (unsafeCrashWith)
 import PureScript.Backend.Chez.Syntax (ChezDefinition(..), ChezExport(..), ChezExpr, ChezImport(..), ChezImportSet(..), ChezLibrary)
 import PureScript.Backend.Chez.Syntax as S
 import PureScript.Backend.Optimizer.Convert (BackendModule, BackendBindingGroup)
-import PureScript.Backend.Optimizer.CoreFn (Ident(..), Literal(..), ModuleName(..), Qualified(..))
+import PureScript.Backend.Optimizer.CoreFn (Ident(..), Literal(..), ModuleName(..))
 import PureScript.Backend.Optimizer.Semantics (NeutralExpr(..))
 import PureScript.Backend.Optimizer.Syntax (BackendAccessor(..), BackendOperator(..), BackendOperator1(..), BackendOperator2(..), BackendOperatorNum(..), BackendOperatorOrd(..), BackendSyntax(..), Pair(..))
 import Safe.Coerce (coerce)
@@ -108,11 +108,8 @@ codegenTopLevelBinding codegenEnv (Tuple (Ident i) n) =
 
 codegenExpr :: CodegenEnv -> NeutralExpr -> ChezExpr
 codegenExpr codegenEnv@{ currentModule } (NeutralExpr s) = case s of
-  Var (Qualified (Just moduleName) (Ident v))
-    | currentModule == moduleName -> S.Identifier v
-    | otherwise -> S.Identifier $ coerce moduleName <> "." <> v
-  Var (Qualified Nothing (Ident v)) ->
-    S.Identifier v
+  Var qi ->
+    S.Identifier $ S.resolve currentModule qi
   Local i l ->
     S.Identifier $ coerce $ S.toChezIdent i l
   Lit l ->
@@ -212,13 +209,8 @@ codegenPrimOp codegenEnv@{ currentModule } = case _ of
         S.List [ S.Identifier "scm:fl-", x' ]
       OpArrayLength ->
         S.List [ S.Identifier "scm:vector-length", x' ]
-      OpIsTag (Qualified Nothing (Ident i)) ->
-        S.app (S.Identifier $ S.recordTypePredicate i) x'
-      OpIsTag (Qualified (Just moduleName) (Ident i))
-        | currentModule == moduleName ->
-            S.app (S.Identifier $ S.recordTypePredicate i) x'
-        | otherwise ->
-            S.app (S.Identifier $ coerce moduleName <> "." <> S.recordTypePredicate i) x'
+      OpIsTag qi ->
+        S.app (S.Identifier $ S.recordTypePredicate $ S.resolve currentModule qi) x'
   Op2 o x y ->
     let
       x' = codegenExpr codegenEnv x
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -11,6 +11,7 @@ import Data.Newtype (unwrap, wrap)
 import Data.Set as Set
 import Data.String.CodeUnits as CodeUnits
 import Data.Tuple (Tuple(..), uncurry)
+import Partial.Unsafe (unsafeCrashWith)
 import PureScript.Backend.Chez.Syntax (ChezDefinition(..), ChezExport(..), ChezExpr, ChezImport(..), ChezImportSet(..), ChezLibrary)
 import PureScript.Backend.Chez.Syntax as S
 import PureScript.Backend.Optimizer.Convert (BackendModule, BackendBindingGroup)
@@ -87,9 +88,7 @@ codegenTopLevelBinding codegenEnv (Tuple (Ident i) n) =
     UncurriedAbs a e ->
       DefineUncurriedFunction i (uncurry S.toChezIdent <$> a) (codegenExpr codegenEnv e)
     CtorDef _ _ _ ss ->
-      case NonEmptyArray.fromArray ss of
-        Nothing -> DefineUncurriedFunction i [] $ codegenExpr codegenEnv n
-        Just xs -> DefineCurriedFunction i xs $ codegenExpr codegenEnv n
+      DefineRecordType i ss
     _ ->
       DefineValue i $ codegenExpr codegenEnv n
 
@@ -126,8 +125,8 @@ codegenExpr codegenEnv@{ currentModule } (NeutralExpr s) = case s of
 
   CtorSaturated _ _ _ _ _ ->
     S.Identifier "ctor-saturated"
-  CtorDef _ _ i ss ->
-    codegenCtorDef i ss
+  CtorDef _ _ _ _ ->
+    unsafeCrashWith "Unreacheable"
 
   LetRec _ _ _ ->
     S.Identifier "let-rec"
@@ -283,17 +282,3 @@ codegenPrimOp codegenEnv = case _ of
           S.List [ S.Identifier "scm:string-append", x', y' ]
         OpStringOrd o' ->
           makeComparison "string" o'
-
-codegenCtorDef :: Ident -> Array String -> ChezExpr
-codegenCtorDef (Ident i) [] =
-  S.List
-    [ S.Identifier "scm:cons"
-    , S.List [ S.Identifier "scm:quote", S.Identifier i ]
-    , S.List [ S.Identifier "scm:quote", S.Identifier "()" ]
-    ]
-codegenCtorDef (Ident i) ss =
-  S.List
-    [ S.Identifier "scm:cons"
-    , S.List [ S.Identifier "scm:quote", S.Identifier i ]
-    , S.List $ [ S.Identifier "scm:vector" ] <> map S.Identifier ss
-    ]
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -17,7 +17,7 @@ import PureScript.Backend.Chez.Syntax as S
 import PureScript.Backend.Optimizer.Convert (BackendModule, BackendBindingGroup)
 import PureScript.Backend.Optimizer.CoreFn (Ident(..), Literal(..), ModuleName(..), Qualified(..))
 import PureScript.Backend.Optimizer.Semantics (NeutralExpr(..))
-import PureScript.Backend.Optimizer.Syntax (BackendOperator(..), BackendOperator1(..), BackendOperator2(..), BackendOperatorNum(..), BackendOperatorOrd(..), BackendSyntax(..), Pair(..))
+import PureScript.Backend.Optimizer.Syntax (BackendAccessor(..), BackendOperator(..), BackendOperator1(..), BackendOperator2(..), BackendOperatorNum(..), BackendOperatorOrd(..), BackendSyntax(..), Pair(..))
 import Safe.Coerce (coerce)
 
 type CodegenEnv =
@@ -131,8 +131,12 @@ codegenExpr codegenEnv@{ currentModule } (NeutralExpr s) = case s of
   UncurriedEffectAbs _ _ ->
     S.Identifier "uncurried-effect-ab"
 
-  Accessor _ _ ->
-    S.Identifier "object-accessor"
+  Accessor _ (GetProp _) ->
+    S.Identifier $ "accessor-get-prop"
+  Accessor _ (GetIndex _) ->
+    S.Identifier $ "accessor-get-index"
+  Accessor e (GetOffset o) ->
+    S.recordAccessor (codegenExpr codegenEnv e) o
   Update _ _ ->
     S.Identifier "object-update"
 
@@ -192,7 +196,7 @@ codegenLiteral codegenEnv = case _ of
   LitRecord r -> S.record $ (map $ codegenExpr codegenEnv) <$> r
 
 codegenPrimOp :: CodegenEnv -> BackendOperator NeutralExpr -> ChezExpr
-codegenPrimOp codegenEnv = case _ of
+codegenPrimOp codegenEnv@{ currentModule } = case _ of
   Op1 o x -> do
     let
       x' = codegenExpr codegenEnv x
@@ -207,8 +211,13 @@ codegenPrimOp codegenEnv = case _ of
         S.List [ S.Identifier "scm:fl-", x' ]
       OpArrayLength ->
         S.List [ S.Identifier "scm:vector-length", x' ]
-      OpIsTag _ ->
-        S.Identifier "op-is-tag"
+      OpIsTag (Qualified Nothing (Ident i)) ->
+        S.app (S.Identifier $ S.recordTypePredicate i) x'
+      OpIsTag (Qualified (Just moduleName) (Ident i))
+        | currentModule == moduleName ->
+            S.app (S.Identifier $ S.recordTypePredicate i) x'
+        | otherwise ->
+            S.app (S.Identifier $ coerce moduleName <> "." <> S.recordTypePredicate i) x'
   Op2 o x y ->
     let
       x' = codegenExpr codegenEnv x
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -78,19 +78,29 @@ codegenTopLevelBindingGroup codegenEnv { recursive, bindings }
   | recursive, Just bindings' <- NonEmptyArray.fromArray bindings =
       [ DefineValue "rtlbg" $ S.Identifier "recursive-top-level-binding-group" ]
   | otherwise =
-      codegenTopLevelBinding codegenEnv <$> bindings
+      Array.concatMap (codegenTopLevelBinding codegenEnv) bindings
 
-codegenTopLevelBinding :: CodegenEnv -> Tuple Ident NeutralExpr -> ChezDefinition
+codegenTopLevelBinding :: CodegenEnv -> Tuple Ident NeutralExpr -> Array ChezDefinition
 codegenTopLevelBinding codegenEnv (Tuple (Ident i) n) =
   case unwrap n of
     Abs a e ->
-      DefineCurriedFunction i (uncurry S.toChezIdent <$> a) (codegenExpr codegenEnv e)
+      [ DefineCurriedFunction i (uncurry S.toChezIdent <$> a) (codegenExpr codegenEnv e) ]
     UncurriedAbs a e ->
-      DefineUncurriedFunction i (uncurry S.toChezIdent <$> a) (codegenExpr codegenEnv e)
+      [ DefineUncurriedFunction i (uncurry S.toChezIdent <$> a) (codegenExpr codegenEnv e) ]
     CtorDef _ _ _ ss ->
-      DefineRecordType i ss
+      case NonEmptyArray.fromArray ss of
+        Nothing -> [ DefineRecordType i [] ]
+        Just xs ->
+          if NEA.length xs == 1
+          then [ DefineRecordType i ss ]
+          else [ DefineRecordType i ss
+               , DefineCurriedFunction i xs
+                 $ S.List
+                 $ Array.cons (S.Identifier $ i <> "*")
+                 (map S.Identifier ss)
+               ]
     _ ->
-      DefineValue i $ codegenExpr codegenEnv n
+      [ DefineValue i $ codegenExpr codegenEnv n ]
 
 codegenExpr :: CodegenEnv -> NeutralExpr -> ChezExpr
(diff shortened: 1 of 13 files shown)


src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -318,3 +322,17 @@ codegenPrimOp codegenEnv = case _ of
...
+    [ S.Identifier "scm:cons"
+    , S.List [ S.Identifier "scm:quote", S.Identifier i ]
+    , S.Identifier "scm:nil"
+    ]
+codegenCtorDef (Ident i) ss =
+  S.List
+    [ S.Identifier "scm:cons"
+    , S.List [ S.Identifier "scm:quote", S.Identifier i ]
+    , S.List $ [ S.Identifier "scm:vector" ] <> map S.Identifier ss
+    ]
```
JordanMartinez, line 338: For constructors with no args, why not just use a symbol?

Per [Section 11.13](http://www.r6rs.org/final/r6rs.pdf):
> Vectors are heterogeneous structures whose elements are indexed by integers. A vector typically occupies **less space than a list of the same length**, and the **average time needed to access a randomly chosen element is typically less for the vector than for the list** (emphasis mine).

For constructors with args, why not use vectors here?

turlando: > For constructors with no args, why not just use a symbol?

The following is based on assumptions that might be completely wrong as I'm not yet very familiar with purscript-backend-optimizer. Please feel free to correct me.

At the discrimination site between different constructors of the same type (OpIsTag), we're only provided with the constructor identifier. This information is not sufficient to choose whether to emit something like `(eq? tag 'Foo)` (for a constructor without args) or `(eq? (car x) 'Foo)` (for a constructor with arguments). Similar considerations apply for the constructor invocation (or lack thereof).

To do what you suggest we might need to keep the CtorMeta for all the types around. I'm not sure if that's a good tradeoff considering that in both `(define Nothing 'Nothing)` and `(define (lambda () 'Nothing))`, `'Nothing` is an interned symbol and no allocation is performed.

> For constructors with args, why not use vectors here?

So you're suggesting to emit `(vector 'Just value0)` instead of `('Just . (vector value0))`? That's a good point! I used that structure because it's quite common in Scheme to tag structures like that, by having a pair whose car is the tag. Maybe we might also want to discuss using records (cf. #84)?

purefunctor: I think constructors may map to records better, since we also get predicate functions for free when checking for OpIsTag—that's my guess at least.

JordanMartinez: @turlando One PR that was merged but not yet released is https://github.com/aristanetworks/purescript-backend-optimizer/pull/44. I was thinking we'd do something similar here. But looking at that PR again, it only works if all the constructors in a sum type have 0 args. If any of them have at least 1, then I'm not sure it will work.

natefaubion: Thats correct, it only optimizes data types that are only enums. I would be shocked if there was a measurable advantage from a runtime/memory perspective since only one constructor is ever allocated, I think it’s mainly that it looks cleaner, is somewhat expected, and was not difficult to implement.

turlando:
Benchmarking the various options considered to encode ADTs real quick (cf. #86).

The encoded data in exam is `Just 1`.

1. Space complexity
   Following the instructions described in [CSUG 9.5 - Debugging, section 3.6: Nested object size and composition](https://cisco.github.io/ChezScheme/csug9.5/debug.html#g21).

   1. Pair of symbol and vector
      Scheme definition: `(cons 'Just (vector 1))`
      ```
      > (compute-size (cons 'Just (vector 1)))
      112

      > (compute-composition (cons 'Just (vector 1)))
      ((pair 1 . 16)
        (symbol 1 . 48)
        (vector 1 . 16)
        (string 1 . 32))
      ```
      I'm not sure where is that string coming from.

   2. Vector
      Scheme definition: `(vector 'Just 1)`
      ```
      > (compute-size (vector 'Just 1))
      112

      > (compute-composition (vector 'Just 1))
      ((symbol 1 . 48) (vector 1 . 32) (string 1 . 32))
      ```

   3. Record
      Scheme definition:
      ```
      (define-record-type (Just Just Just?) (fields value0))
      (Just 1)
      ```
      ```
      > (define-record-type (Just Just Just?) (fields value0))

      ;; As per documentation, we have to force the record type definition
      ;; objects into GC generation 1.
      > (collect 1 1)

      > (compute-size (Just 1) 0)
      16

      > (compute-composition (Just 1) 0)
      ((#<record type Just> 1 . 16))
      ```

2. Deduplication
   Ideally we might want a single `Nothing` to ever float around in memory. In that regard records (for trivial cases at least) behave as interned symbols.

   ```
   > (define-record-type (Nothing Nothing Nothing?) (fields))
   > (define n1 (Nothing))
   > (define n2 (Nothing))
   > (eq? n1 n2)
   #t

   > (define c1 (cons 'Nothing '()))
   > (define c2 (cons 'Nothing '()))
   > (eq? c1 c2)
   #f

   > (define v1 (vector 'Nothing))
   > (define v2 (vector 'Nothing))
   > (eq? v1 v2)
   #f

   > (define s1 'Nothing)
   > (define s2 'Nothing)
   > (eq? s1 s2)
   #t
   ```

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -126,8 +124,10 @@ codegenExpr codegenEnv@{ currentModule } (NeutralExpr s) = case s of
 
   CtorSaturated _ _ _ _ _ ->
     S.Identifier "ctor-saturated"
-  CtorDef _ _ i ss ->
-    codegenCtorDef i ss
+  CtorDef _ _ _ _ ->
+    -- FIXME: this case is already completely handled in codegenTopLevelBinding,
+    -- not sure what to do with this.
+    S.Identifier "CtorDef"
```
purefunctor, line 130: I think an `unsafeCrashWith` would be apt here

JordanMartinez:
The codegen we use for data constructors needs to work with the [`Accessor constructor's BackendAccessor`](https://github.com/aristanetworks/purescript-backend-optimizer/blob/main/src/PureScript/Backend/Optimizer/Syntax.purs#L22)'s constructor [`GetOffset`](https://github.com/aristanetworks/purescript-backend-optimizer/blob/main/src/PureScript/Backend/Optimizer/Syntax.purs#L54). 

```purs
data Foo = Foo Int

-- equivalent BackendSyntax (I believe):
--   `App fooValue (Accessor $ GetOffset 0)`
getOffset0 :: Foo -> Int
getOffset0 (Foo i) = i
```

It seems the record accessors are type-specific. So, can ADTs be encoded using `define-record-type`? I don't know we have information about the type to which `foo` corresponds available when a `Accessor $ GetOffset 2` appears.

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -24,18 +25,20 @@ type CodegenEnv =
...
-
     definitions :: Array ChezDefinition
     definitions = Array.concat $
       codegenTopLevelBindingGroup codegenEnv <$> bindings
 
+    exports' :: Array ChezExport
+    exports' = map ExportIdentifier
+      $ Array.concatMap S.definitionIdentifiers definitions
+          <> map coerce (Array.fromFoldable foreign_)
+
```
purefunctor, line 41: Bikeshedding question, should exports be ordered alphabetically?

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -123,7 +138,7 @@ codegenExpr codegenEnv@{ currentModule } (NeutralExpr s) = case s of
   CtorSaturated _ _ _ _ _ ->
     S.Identifier "ctor-saturated"
   CtorDef _ _ _ _ ->
-    S.Identifier "ctor-def"
+    unsafeCrashWith "Unreacheable"
```
purefunctor, line 141:
```suggestion
    unsafeCrashWith "codegenExpr:CtorDef - handled by codegenTopLevelBinding!"
```

JordanMartinez:
If our goal is to get a compiled output for Chez Scheme as quickly as possible, then we should just use vectors for now. I don't think that will force us to change any FFI in the future as any constructors for ADTs will be passed into the FFI as an argument. And then we can reconsider this in the future once we've verified other things work.

That being said, we could encode ADTs as records where one of the fields is the constructor's tag. Then, the type is just number of fields plus 1:
```scheme
; defined in our runtime module
(define-record-type (Record0$ Record0* Record0?) (fields tag))
(define-record-type (Record1$ Record1* Record1?) (fields tag value1))
(define-record-type (Record2$ Record2* Record2?) (fields tag value1 value2))
; data Foo = Foo
(define Foo (Record0* "Foo")
; data Bar = Bar 1 | Baz 2 3
(define Bar (lambda (i) (Record1* "Bar" 1))
(define Baz (lamda (i1) (lambda (i2) (Record2* "Baz" i1 i2))))
```

Then, maybe using that `record-rtd` and `record-accessor` approach would work? But TSPLv4 does say in [Section 9.2](https://scheme.com/tspl4/records.html#./records:h2):
> The procedural interface is more flexible than the syntactic interface, but this flexibility can lead to less readable and efficient programs, so programmers should use the syntactic interface whenever it suffices.

turlando:
> That being said, we could encode ADTs as records where one of the fields is the constructor's tag. Then, the type is just number of fields plus 1:
> 
> ```scheme
> ; defined in our runtime module
> (define-record-type (Record0$ Record0* Record0?) (fields tag))
> (define-record-type (Record1$ Record1* Record1?) (fields tag value1))
> (define-record-type (Record2$ Record2* Record2?) (fields tag value1 value2))
> ; data Foo = Foo
> (define Foo (Record0* "Foo")
> ; data Bar = Bar 1 | Baz 2 3
> (define Bar (lambda (i) (Record1* "Bar" 1))
> (define Baz (lambda (i1) (lambda (i2) (Record2* "Baz" i1 i2))))
> ```

Unfortunately that would take away the nice singleton property.

```
> (define-record-type (Nothing$ Nothing* Nothing?) (fields tag))
> (define n0 (Nothing* 'Nothing))
> (define n1 (Nothing* 'Nothing))
> (eq? n0 n1)
#f
```

Moreover I think that encoding only the tag is not sufficient. If the goal is to construct an accessor symbol (e.g. `Foo-value0`), then we might need a FQN.

> Then, maybe using that `record-rtd` and `record-accessor` approach would work? But TSPLv4 does say in [Section 9.2](https://scheme.com/tspl4/records.html#./records:h2):
> 
> > The procedural interface is more flexible than the syntactic interface, but this flexibility can lead to less readable and efficient programs, so programmers should use the syntactic interface whenever it suffices.

I'm not sure whether that statements applies only to `make-record-type-descriptor`. Consider that `record-rtd` is mentioned in the following section "Inspection", whose introduction says:

>This section describes various procedures for asking questions about or extracting information from record-type descriptors (rtds). It also describes the record-rtd procedure, with which the rtd of a non-opaque record instance may be extracted, allowing the record type of the instance to be inspected and, via record accessors and mutators generated from the rtd, the record itself to be inspected or modified. This is a powerful feature that permits the coding of portable record printers and inspectors.

Summing it up, the code we would have to produce is `((record accessor (record-rtd n) x) n)`, where n is a record instantiation and x is the field offset.

JordanMartinez:
> Unfortunately that would take away the nice singleton property.

Why does that matter? Is there some performance benefits that we give up by dropping this?

JordanMartinez:
I've got a PR up that adds type and constructor information to `GetOffset` (and renames that to `GetCtorField` here: https://github.com/aristanetworks/purescript-backend-optimizer/pull/45. I think with this additional information, we should be able to use the original `define-record-type` encoding.

src/PureScript/Backend/Chez/Convert.purs
turlando: > For constructors with no args, why not just use a symbol?

[bf70fee](https://github.com/purescm/purescm/pull/85/commits/bf70fee3ccb89f7dcfd5ce5726382397c8581e08) implements it.

JordanMartinez:
Tests that we can add to this PR:
```purs
data Foo = Foo Int Int Int

-- a constructor that is partially applied to its arguments.
x = Foo 1
y = Foo 1 1
z = Foo 1 1 1

-- more pattern matching tests
data NoArgs = NoArgs

test1 = case _ of
  NoArgs -> true

data HasArgs = HasArgs Int Int Int

test2 = case _ of
  HasArgs i1 _ _ -> i1
  
test3 = case _ of
  HasArgs i1 i2 i3 
    | i1 < i3 -> i1
    | otherwise -> i2

data SumWithArgs
  = First Int
  | Last Int

test4 = case _ of
  First i -> i
  Last i -> i

test5 :: Partial => SumWithArgs -> Int
test5 = case _ of
  First i -> i
```

JordanMartinez:
What is the snapshot output when a constructor is not exported?

src/PureScript/Backend/Chez/Syntax.purs
```diff
@@ -103,10 +103,14 @@ definitionIdentifiers :: ChezDefinition -> Prim.Array Prim.String
...
 definitionIdentifiers (DefineUncurriedFunction i _ _) = [ i ]
-definitionIdentifiers (DefineRecordType i _) =
+definitionIdentifiers (DefineRecordType i []) =
   [ recordTypeUncurriedConstructor i
   , recordTypePredicate i
   ]
+definitionIdentifiers (DefineRecordType i fields) =
+  [ recordTypeUncurriedConstructor i
+  , recordTypePredicate i
+  ] <> map (recordTypeAccessor i) fields
```
purefunctor: I think matching against an empty array is redundant here? Unless there's some extra code you want to add when it's empty.
```suggestion
definitionIdentifiers (DefineRecordType i fields) =
  [ recordTypeUncurriedConstructor i
  , recordTypePredicate i
  ] <> map (recordTypeAccessor i) fields
```

src/PureScript/Backend/Chez/Syntax.purs
```diff
@@ -98,6 +99,25 @@ data ChezExpr
...
+definitionIdentifiers (DefineCurriedFunction i _ _) = [ i ]
+definitionIdentifiers (DefineUncurriedFunction i _ _) = [ i ]
+definitionIdentifiers (DefineRecordType i []) =
+  [ recordTypeUncurriedConstructor i
+  , recordTypePredicate i
+  ]
+definitionIdentifiers (DefineRecordType i fields) =
+  [ recordTypeUncurriedConstructor i
+  , recordTypePredicate i
+  ] <> map (recordTypeAccessor i) fields
```
purefunctor, line 113: I think matching against an empty array is redundant here? Unless there's some extra code you want to add when it's empty. (Resubmitted this comment because somehow, I might've commented on an earlier commit)
```suggestion
definitionIdentifiers (DefineRecordType i []) =
  [ recordTypeUncurriedConstructor i
  , recordTypePredicate i
  ]
definitionIdentifiers (DefineRecordType i fields) =
  [ recordTypeUncurriedConstructor i
  , recordTypePredicate i
  ] <> map (recordTypeAccessor i) fields
```

turlando:
> What is the snapshot output when a constructor is not exported?

We were always exporting all the definitions. Should be fixed in [e16e190](https://github.com/purescm/purescm/pull/85/commits/e16e1905b30c66ebecd41a79fac0db28a7e20aab). I'd love to test it but between inlining and dead code elimination I'm not sure how to do it.

I'm not particularly happy with the `DefinitionTranslation` name.

JordanMartinez:
Could you add a snapshot for that? For example

```purs
module Snapshot.Exports
  ( TypeOnly
  , TypeAndCtor
  ) where

data TypeOnly = Ctor

data TypeAndCtor = TypeAndCtor
```

turlando:
Thanks @JordanMartinez, I was definitely overthinking it. Added that test case to the last commit and looks like it's working as expected.

test-snapshots/snapshots-input/Snapshot.Exports.purs
```diff
@@ -0,0 +1,8 @@
+module Snapshot.Exports
+  ( TypeOnly
+  , TypeAndCtor
```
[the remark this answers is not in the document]
turlando: Uhm, actually I was thinking that we wanted to test that it would *not* export the constructors, as we already know that exporting them works.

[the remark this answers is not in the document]
JordanMartinez: >  as we already know that exporting them works.

Right, but there are other things we can export besides constructors. And that may have implications for later things when type classes get involved.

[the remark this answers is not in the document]
JordanMartinez: Granted, at this point, I think there's been a bit of feature creep as now we're hitting a lot more pieces of code that are now relevant because of ADT conversion being done in this PR.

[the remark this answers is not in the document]
turlando: Right, I will update the test then. Should we keep snapshot to test that it will non export everything blindly as well?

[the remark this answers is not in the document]
JordanMartinez: > Right, I will update the test then. Should we keep snapshot to test that it will non export everything blindly as well?

I'm not sure what you mean.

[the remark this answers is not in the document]
turlando: Sorry, I think I misread your first comment. So you're suggesting to change the snapshot to

```haskell
module Snapshot.Exports
  ( TypeOnly
  , TypeAndCtor(..)
  )
...
```

Right?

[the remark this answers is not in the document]
JordanMartinez: Yeah.

JordanMartinez:
If this PR is ready for a final review, could you take it out of the 'draft' state?

test-snapshots/snapshots-output/Snapshot.Constructor.ss
```diff
@@ -0,0 +1,47 @@
+#!r6rs
+#!chezscheme
+(library
+  (Snapshot.Constructor lib)
+  (export
+    Just*
+    Just-value0
+    Just?
```
JordanMartinez, line 8: Something to consider for later (perhaps?) but does it matter to the Chez Scheme code that the `Just` type isn't exported?

test-snapshots/snapshots-output/Snapshot.Constructor.ss
```diff
@@ -0,0 +1,47 @@
...
+    Node*
+    Node-value0
+    Node-value1
+    Node-value2
+    Node?
+    Nothing
+    Nothing?)
+  (import
+    (prefix (chezscheme) scm:)
+    (prefix (_Chez_Runtime lib) rt:))
```
JordanMartinez, line 21: Something to think about later, but if we aren't using anything from the runtime library, should we import it in the generated output?

test-snapshots/snapshots-output/Snapshot.ConstructorAccessor.ss
```diff
@@ -0,0 +1,97 @@
+#!r6rs
+#!chezscheme
+(library
+  (Snapshot.ConstructorAccessor lib)
+  (export
+    First*
+    First-value0
+    First?
```
JordanMartinez, line 8: At first, I thought the uncurried version of `First*` wasn't being exported here. Then I realized that `First` only takes a single argument. 

In cases like this (a ctor that takes 1 argument), should the constructor name be `First`? 

We could always export two constructors (a curried and uncurried version), but the only reason why that might be desirable is if one wanted to use the uncurried constructor directly in FFI files, which I think would be a bad practice to encourage.

test-snapshots/snapshots-output/Snapshot.ConstructorAccessor.ss
```diff
@@ -0,0 +1,97 @@
...
+    (scm:fields (scm:immutable value0 Foo-value0) (scm:immutable value1 Foo-value1) (scm:immutable value2 Foo-value2)))
+
+  (scm:define Foo
+    (scm:lambda (value0)
+      (scm:lambda (value1)
+        (scm:lambda (value2)
+          (Foo* value0 value1 value2)))))
+
+  (scm:define z
+    (Foo* 1 1 1))
```
JordanMartinez, line 70: Just highlighting here that the uncurried function is used when all args are provided to the constructor. :heavy_check_mark:

test-snapshots/snapshots-output/Snapshot.ConstructorAccessor.ss
```diff
@@ -0,0 +1,97 @@
...
+          (HasArgs* value0 value1 value2)))))
+
+  (scm:define-record-type (Foo$ Foo* Foo?)
+    (scm:fields (scm:immutable value0 Foo-value0) (scm:immutable value1 Foo-value1) (scm:immutable value2 Foo-value2)))
+
+  (scm:define Foo
+    (scm:lambda (value0)
+      (scm:lambda (value1)
+        (scm:lambda (value2)
+          (Foo* value0 value1 value2)))))
```
JordanMartinez, line 67: Seems like the order of the outputted definitions don't match the source order. I would expect the order to be:
- `Foo` (e.g. `Foo$`, `Foo`)
- `NoArgs` (e.g. `NoArgs`, `NoArgs?`)
- `HasArgs`
- `First`
- `Last`

test-snapshots/snapshots-output/Snapshot.ConstructorAccessor.ss
```diff
@@ -0,0 +1,97 @@
...
+
+  (scm:define z
+    (Foo* 1 1 1))
+
+  (scm:define y
+    ((Foo 1) 1))
+
+  (scm:define x
+    (Foo 1))
```
JordanMartinez, line 76: Similar situation here in regards to source order.

test-snapshots/snapshots-output/Snapshot.Exports.ss
```diff
@@ -0,0 +1,17 @@
...
+  (import
+    (prefix (chezscheme) scm:)
+    (prefix (_Chez_Runtime lib) rt:))
+
+  (scm:define TypeAndCtor
+    (scm:quote TypeAndCtor))
+
+  (scm:define TypeAndCtor?
+    (scm:lambda (v)
+      (scm:eq? (scm:quote TypeAndCtor) v))))
```
JordanMartinez, line 17: Unexpectedly, `TypeOnly` isn't defined here. Maybe `purescript-backend-optimizer` removes it because it's unused? If that is the case, we could stop that from happening by adding a function to the snapshot that uses a value of that type and exports that function. So, something like
```purs
module Snapshot.Exports
  ( TypeOnly
  , ensureTypeOnlyIsUsed
  , TypeAndCtor(..)
  ) where

data TypeOnly = Ctor String

ensureTypeOnlyIsUsed :: TypeOnly -> String
ensureTypeOnlyIsUsed (Ctor s) = s

data TypeAndCtor = TypeAndCtor
```

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -10,31 +10,50 @@ import Data.Maybe (Maybe(..))
...
+
+    exports' :: Array ChezExport
+    exports' = map ExportIdentifier
+      $ Array.sort
+      $ Array.concatMap S.definitionIdentifiers definitions'
+          <> map coerce (Array.fromFoldable foreign_)
+      where
+      translations' :: Array DefinitionTranslation
+      translations' =
+        Array.filter (\t -> Set.member t.pursIdent exports) translations
```
JordanMartinez, line 53: Could you clarify why you didn't add `exports` to the `codegenEnv`, and then determine whether a definition should be exported in `codegenToplevelBinding`?

I think that would clean up this code.

test-snapshots/snapshots-output/Snapshot.Constructor.ss
JordanMartinez: Tracked in https://github.com/purescm/purescm/issues/92

test-snapshots/snapshots-output/Snapshot.Constructor.ss
JordanMartinez: Tracked in https://github.com/purescm/purescm/issues/93

test-snapshots/snapshots-output/Snapshot.ConstructorAccessor.ss
turlando: Fixed in [217e8dc](https://github.com/purescm/purescm/pull/85/commits/217e8dc6c1ebddbfd06b9fb87e691a6e4b534075), but I'm not very happy with the overall state of the code. I feel like I'm spreading too much stuff across `codegenTopLevelBinding'`, `printDefinition` and `definitionIdentifiers`.

test-snapshots/snapshots-output/Snapshot.Exports.ss
turlando: And that's absolutely broken

```
#!r6rs
#!chezscheme
(library
  (Snapshot.Exports lib)
  (export
    TypeAndCtor
    TypeAndCtor?
    ensureTypeOnlyIsUsed)
  (import
    (prefix (chezscheme) scm:)
    (prefix (_Chez_Runtime lib) rt:))

  (scm:define TypeAndCtor
    (scm:quote TypeAndCtor))

  (scm:define TypeAndCtor?
    (scm:lambda (v)
      (scm:eq? (scm:quote TypeAndCtor) v)))

  (scm:define ensureTypeOnlyIsUsed
    (scm:lambda (v0)
      (Ctor-value0 v0))))
```

The weird thing is that if I try to print the idents of the bindings that are supposed to be translates with

```
bindingsIdents :: Array String                                                                                                                                           
bindingsIdents = map (coerce <<< Tuple.fst) $ Array.concatMap (_.bindings) bindings
```

in `codegenModule`, and somehere later I `trace` that value, I obtain the following

```
[38 of 42] Building Snapshot.Exports
["TypeAndCtor","ensureTypeOnlyIsUsed"]
```

There is no trace (no pun intended) of TypeOnly's Ctor. Maybe it's trying to inline it but we fail to generate the code for it? Or is it a bug in backend-optimizer? I'm confused...

test-snapshots/snapshots-output/Snapshot.ConstructorAccessor.ss
turlando: The ordering is provided by BackendModule.bindings. Tracing the idents as follows

```haskell
bindingsIdents :: Array String                                                                                                                                           
bindingsIdents = map (coerce <<< Tuple.fst) $ Array.concatMap (_.bindings) bindings
```

Produces

```
[37 of 42] Building Snapshot.ConstructorAccessor                                                                                                                             
["First","Last","NoArgs","HasArgs","Foo","z","y","x","test5","test4","test3","test2","test1"]
```

I'm not sure how we can get the source code order.

JordanMartinez: I've opened https://github.com/aristanetworks/purescript-backend-optimizer/issues/46 upstream. Maybe it can be fixed there? Maybe not?

test-snapshots/snapshots-output/Snapshot.Exports.ss
JordanMartinez: I think this is a bug in backend-optimizer... `bindings` come from `usedBindings.value`: https://github.com/aristanetworks/purescript-backend-optimizer/blob/main/src/PureScript/Backend/Optimizer/Convert.purs#L159-L191. Notice how only the identifier is checked? In the above situation, the type and constructor share the same name. Because the constructor isn't exported, that code may also be removing the type as well.

turlando:
As discussed with @JordanMartinez I'm closing this PR as it's getting way too messy.

PR #88, opened 2023-03-25
Implement code generation for Effect

purefunctor:
This also changes code generation for `Let` such that a `let*` is generated for nested `Let`s in the backend syntax. Similarly, the `Effect` variants are compiled into a `let*` as seen in the snapshot test. This also adds support for running snapshot tests with `main` declarations.


purefunctor:
I'll write code to get `assert` working later today, such that we can test runtime behaviour as well.

JordanMartinez (approved):

purefunctor:
I'll complete the code for runtime tests then we can merge

purefunctor:
@JordanMartinez I've ported `assert` over to Scheme, however full programs can't be ran yet, with a blocker right now being ADTs. We can either wait for #85 or merge this and create a new PR.

test-snapshots/packages.dhall
```diff
@@ -6,10 +6,25 @@ in  upstream
...
   with partial =
     { dependencies = [] : List Text
     , repo = "https://github.com/purescm/purescript-partial.git"
-    , version = "d8b6daf068e4aab0ad85a42d0f890ed37f416923"
+    , version = "v4.0.0-scm"
+    }
+  with effect =
+    { dependencies = [ "prelude" ]
+    , repo = "https://github.com/purescm/purescript-effect.git"
+    , version = "v4.0.0-scm"
```
f-f, line 19: I'd advise against pointing spago to branches, because new commits won't be picked up over time. Commit hashes are the most reliable.

PR #94, opened 2023-03-29
Switch to the new spago

f-f:
It's included in the nix-shell, but if you're not using that you might want to install it with `npm -g install spago@next`


turlando:
I think it's better to have the CI and the nix shell synchronized and we should have the same Chez version on both.

f-f:
We can use the nix-shell in CI, but that lengthens the CI turnaround (as loading nix stuff is slow, or at least it used to be). We can do that if everyone is ok with that, otherwise I am open to suggestions on how to sync the versions.

f-f:
We could also switch to flakes as [we have in the registry-dev repo](https://github.com/purescript/registry-dev/blob/master/.github/workflows/main.yml), but that can be done in a separate PR.

turlando:
> @turlando switched to nix in CI.

Great! Thanks a lot really! Let's see how the CI times compare. Also: maybe we can make the shell pure?

f-f:
> maybe we can make the shell pure?

What do you mean?

turlando:
Looks good to me, but could you please upstream your changes to easy-purescript-nix?

turlando (approved):

f-f:
Changes can't be upstreamed until we release the new spago at least in alpha, and even then we'd need to agree with the maintainer on what's the best way to make it available (e.g. I'd prefer to wholesale replace the old spago, but others might have other preferences). This is all to say that it's not straightforward, so won't happen for some weeks at least (even though we are close to an alpha of the new Spago).
In any case, I am tracking the upstream, so the changes in my fork are only additive.

test/Main.purs
```diff
@@ -84,13 +84,13 @@ main = do
...
-  spawnFromParent "spago" [ "build", "-u", "-g corefn" ]
+  spawnFromParent "spago" [ "build", "--purs-args", "-g corefn" ]
   snapshotDir <- liftEffect Process.cwd
-  snapshotPaths <- expandGlobs (Path.concat [ snapshotDir, "snapshots-input" ])
+  snapshotPaths <- expandGlobs (Path.concat [ snapshotDir, "src", "snapshots-input" ])
     [ "Snapshot.*.purs" ]
   schemeBin <- getSchemeBinary
   outputRef <- liftEffect $ Ref.new Map.empty
-  let snapshotsOut = Path.concat [ snapshotDir, "snapshots-output" ]
+  let snapshotsOut = Path.concat [ snapshotDir, "src", "snapshots-output" ]
```
JordanMartinez, line 93: Just curious but does `src` need to be here? I think the extra directory is a bit unnecessary.

f-f: The purescript files need to be in `src` for Spago to consider it a project (this is a new constraint in the new Spago), but the output files do not need to be there as well. I just moved them together with the PureScript sources.

f-f: The registry has this constraint and Spago just adopts it. The constraint was adopted in the Registry for backwards compatibility with the current situation (where all our tooling expects things to be in `src`) and for consistency, so one does not have to look up a config to find the sources for a given package.

PR #97, opened 2023-03-29
Implement data constructors codegen

turlando:
This PR is a rework of #85.

Constructors are implemented in the following way:
    
  - If their arity is zero, the structure representing the constructor
    is a simple symbol. E.g. if the constructor is Foo, the symbol is
    'Foo. A constructing function with the same name as the constructor
    is also produced alongside a predicate function whose name has a
    trailing question mark (respectively Foo and Foo?).
  - If the arity is one, Scheme records are used. The record type name
    is the constructor name with a trailing dollar character (e.g. Foo$).
    Constructing and predicate functions are provided by the
    define-record-type form and follow the same conventions as
    described in the previous point. The form takes also care of
    producing accessor functions whose name is Ctor-valueN, where Ctor
    is the constructor name and N is the field offset (e.g. Foo-value0).
  - For constructors with arity greater than one, almost everything
    described in the previous point applies. The differences are that
    the constructor has a trailing asterisk character (e.g. Foo*) and a
    curried constructing function is provided, whose name is the same as
    the constructor name.

Moreover codegen for CtorSaturated, OpIsTag and Accessor with
CtorGetField is also handled.

### DIFF
--- a/packages.dhall
+++ b/packages.dhall
@@ -49,8 +49,8 @@ let additions =
           , "unsafe-coerce"
           ]
         , repo =
-            "https://github.com/aristanetworks/purescript-backend-optimizer"
-        , version = "purs-backend-es-v1.3.2"
+            "https://github.com/JordanMartinez/purescript-backend-optimizer"
+        , version = "add-ty-con-info"
         }
       , node-glob-basic =
         { dependencies =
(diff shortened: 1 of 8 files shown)


JordanMartinez (approved):

JordanMartinez:
Thanks for cleaning up commit history.

Issue #65, opened 2021-10-14
Fix Char and String literals emission


f-f:
PureScript uses UTF-8 strings, and we don't support that right now. Instead we are emitting things JS-specific like `\x0a` or `\uXXXX`.
We think that the issue preventing us from emitting proper unicode is somewhere in the `FromJSON` instance for the `CoreFn`, or the `prettyPrintStringJS` that we call when emitting Scheme.

f-f:
We'll need a test to see if this applies to the new implementation 😄

turlando:
According to [TSPL, Formal Syntax](https://scheme.com/tspl4/grammar.html).

---

A character object is written as the prefix #\ followed by a single character, a character name, or a sequence of characters specifying a Unicode scalar value.

```
<character>        --> #\ <any character> | #\ <character name> | #\x <hex scalar value>
<character name>   --> alarm | backspace | delete | esc | linefeed | newline | page | return | space | tab | vtab
<hex scalar value> --> <digit 16>+
```

The named characters correspond to the Unicode characters alarm (Unicode scalar value 7, i.e., U+0007), backspace (U+0008), delete (U+007F), esc (U+001B), linefeed (U+000A; same as newline), newline (U+000A), page (U+000C), return (U+000D), space (U+0020), tab (U+0009) and vertical tab (U+000B).

A hex scalar value represents a Unicode scalar value n, 0 <= n <= D800_16 or E000_16 <= n <= 10FFF_16. The <digit 16> nonterminal is defined under Numbers below.

A #\ prefix followed by a character name is always interpreted as a named character, e.g., #\page is treated as #\page rather than #\p followed by the symbol age. Characters must also be delimited, as described above, so that #\pager is treated as a syntax error rather than as the character #\p followed by the symbol ager or the character #\page followed by the symbol r.

Case is significant in the syntax of character objects, except within a hex scalar value. 

---

A string is written as a sequence of string elements enclosed in string quotes (double quotes). Any character other than a string quote or backslash can appear as a string element. A string element can also consist of a backslash followed by a single character, a backslash followed by sequence of characters specifying a Unicode scalar value, or a backslash followed by sequence of intraline whitespace characters that includes a single line ending.

```
<string>         --> " <string character>* "
<string element> --> <any character except " or \>
                   | \" | \\ | \a | \b | \f | \n | \r | \t | \v
                   | \x <hex scalar value> ;
                   | \ <intraline whitespace>* <line ending> <intraline whitespace>*
```

[...]

PR #101, opened 2023-03-30
Fix string literals codegen

turlando:
See #65.

### DIFF
--- a/src/PureScript/Backend/Chez/Syntax.purs
+++ b/src/PureScript/Backend/Chez/Syntax.purs
@@ -10,8 +10,12 @@ import Data.Array.NonEmpty as NEA
 import Data.Array.NonEmpty as NonEmptyArray
 import Data.Maybe (Maybe(..), maybe)
 import Data.Newtype (class Newtype, un)
+import Data.String.Regex as R
+import Data.String.Regex.Unsafe as R.Unsafe
+import Data.String.Regex.Flags as R.Flags
 import Dodo (Doc)
 import Dodo as D
+import Partial.Unsafe (unsafeCrashWith)
 import Prim as Prim
 import PureScript.Backend.Chez.Constants (libChezSchemePrefix, scmPrefixed)
 import PureScript.Backend.Optimizer.CoreFn (Ident(..), ModuleName(..), Prop(..), Qualified(..))
@@ -344,6 +348,23 @@ toChezIdent :: Maybe Ident -> Level -> Prim.String
 toChezIdent i (Level l) = case i of
   Just (Ident i') -> i' <> show l
   Nothing -> "_" <> show l
+
+jsonToChezString :: Prim.String -> Prim.String
+jsonToChezString str = unicodeReplace str
+  where
+  unicodeRegex :: R.Regex
+  unicodeRegex = R.Unsafe.unsafeRegex """\\u([A-F\d]{4})""" R.Flags.global
+
+  unicodeReplace :: Prim.String -> Prim.String
+  unicodeReplace s = R.replace' unicodeRegex unicodeReplaceMatch s
+
+  unicodeReplaceMatch
+    :: Prim.String
+    -> Prim.Array (Maybe Prim.String)
+    -> Prim.String
+  unicodeReplaceMatch _ = case _ of
+    [ (Just x) ] -> "\\x" <> x <> ";"
+    _ -> unsafeCrashWith "Error matching at unicodeReplaceMatch in jsonToChezString"
 
 --
 
--- a/test-snapshots/src/snapshots-input/Snapshot.Literals.String.purs
+++ b/test-snapshots/src/snapshots-input/Snapshot.Literals.String.purs
@@ -1,0 +1,79 @@
+module Snapshot.Literals.String where
+
+block1 :: String
+block1 = """block"""
+
+block2 :: String
+block2 = """
+foo
+bar
+"""
+
+block3 :: String
+block3 = """foo
+bar"""
+
+block4 :: String
+block4 = """foo
+bar
+baz"""
+
+block5 :: String
+block5 = """foo
+bar
+baz
+"""
+
+---
+
+escape1 :: String
+escape1 = "\t"
+
+escape2 :: String
+escape2 = "\n"
+
+escape3 :: String
+escape3 = "\r"
+
+escape4 :: String
+escape4 = "\r"
+
+escape5 :: String
+escape5 = "\r\n"
+
+escape6 :: String
+escape6 = "\\"
+
+escape7 :: String
+escape7 = "\""
+
+escape8 :: String
+escape8 = "\'"
+
+---
+
+-- Lowest valid code point
+unicode1 :: String
+unicode1 = "\x0"
+
+unicode2 :: String
+unicode2 = "\x10ffff"
+
+-- Emoji
+unicode3 :: String
+unicode3 = "\x2713"
+
+-- Unicode literal with leading zero. Should be equal to unicode3.
+unicode4 :: String
+unicode4 = "\x02713"
+
+-- Multiple unicode literals in a single string
+unicode5 :: String
+unicode5 = "Foo \x2713 Bar \x2713 \n Baz \x2713"
+
+unicode6 :: String
+unicode6 = "2342\x2713"
+
+-- String with ^H
+unicode7 :: String
+unicode7 = "Foo \x8\x8\x8\x8 Bar"
--- a/spago.yaml
+++ b/spago.yaml
@@ -44,7 +44,7 @@ package:
       - refs
 workspace:
   package_set:
-    registry: 19.1.0
+    registry: 20.0.1
   extra_packages:
     backend-optimizer:
       git: https://github.com/JordanMartinez/purescript-backend-optimizer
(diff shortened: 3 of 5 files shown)


test-snapshots/src/snapshots-input/Snapshot.Literals.String.purs
```diff
@@ -0,0 +1,86 @@
...
+
+-- unicode2 :: String
+-- unicode2 = "\x10ffff"
+--
+-- The above definition causes the build to fail.
+--
+-- [43 of 43] Building Snapshot.Literals.String
+-- file:///home/u/purescm/output/Effect.Aff/foreign.js:530
+--                 throw util.fromLeft(step);
+--                 ^
```
JordanMartinez, line 66: If you change [this line](https://github.com/purescm/purescm/blob/master/test/Main.purs#L136) to
```purs
hasMain <- either (unsafeCrashWith <<< append "canRunMain: parse failure - ") pure $
  canRunMain originalFileSourceCode
```
what does it output?

turlando: It's hitting the ParseFailed case.

f-f: Let's change the `Test.Main` to not have that `unsafeCrash` so we can always know what's the relevant error.

JordanMartinez: I should have printed the full error when I added this initially...
```
Error: canRunMain: parse failure

Could not parse file: 
position: { column: 20, line: 2 }
error: Character escape out of range \10ffff

Comment (JordanMartinez) [reply test-snapshots/src/snapshots-input/Snapshot.Literals.String.purs:67]: Pretty sure this is a bug in `cst-parser` because it's not using [`enums@v6.01`](https://github.com/purescript/purescript-enums/releases). The `enums` fix was [released on 2022-08-16](https://github.com/purescript/purescript-enums/releases) but `cst-parser` is using a [package set that was released on 2022-05-07](https://github.com/natefaubion/purescript-language-cst-parser/blob/main/packages.dhall#L2)

Comment (JordanMartinez) [reply test-snapshots/src/snapshots-input/Snapshot.Literals.String.purs:67]: I could be wrong, but that would be my first guess.

Comment (natefaubion) [reply test-snapshots/src/snapshots-input/Snapshot.Literals.String.purs:67]: The CST Parser uses `Char.fromCharCode`, so if it's a bug, it's a bug there.
https://github.com/natefaubion/purescript-language-cst-parser/blob/13a2437571b60c2c861bbd5b0d07850b4cc82e94/src/PureScript/CST/Lexer.purs#L527

Comment (natefaubion) [reply test-snapshots/src/snapshots-input/Snapshot.Literals.String.purs:67]: > The enums fix was [released on 2022-08-16](https://github.com/purescript/purescript-enums/releases) but cst-parser is using a [package set that was released on 2022-05-07](https://github.com/natefaubion/purescript-language-cst-parser/blob/main/packages.dhall#L2)

It wouldn't matter what cst-parser is using internally. What matters is the version you are using in this project.

Comment (JordanMartinez) [reply test-snapshots/src/snapshots-input/Snapshot.Literals.String.purs:67]: I think `Char.fromCharCode` is buggy: https://try.purescript.org/?code=FAWw9gJgrgNgpgAgLIEMCWA7BB3AFnAJzmGDRAAcwCAXBABSJigmNIqtoBEVqUA6VAE8ARogAUQ0WL58AlABoEAMwJgQkuIpVqAUlADO1RSBQi4stpRoIAokqVwAxrTF2Hzi2Su03T6nwBhGBR9fUCwDH0weAQxGDAAc092awAVAkE6KCIAZUcCNHJaEIR0zOy4PIKiyw4Ebl5A3BQCBBKA5oJa6wb%2BAEkMYv0EAeoSE0wEAC4p23s-BABVDDQxiawAXlKMrNz8wv8iDBZWjYAeM%2B3yver-bFXcAIiomIgwYAQEeISEABI28jkODHBAAIgAjAAGABisOhbWGLEcZBQMFmnwxmIxoL%2BCH0uDA2Fxoz42hAOWoBQwCQAgsMSfgAB4oJEomBgqH2eygj5fRK4lCA4EQMEdFp8ahgMUEJ4sWKS8jTWbS2SzHH-fGE3F6Qy46USqWdWWIBW874CoUg0GcNDzI6OODorHOl3O9V4glE-7gn0AFh94IQAFoEAA2ACs4YAzOHeUA

```
10FFFF as decimal:             (Just 1114111)
Char.toCharCode (top :: Char): (Just 65535)
Difference:                          1048576
```

Comment (JordanMartinez) [reply test-snapshots/src/snapshots-input/Snapshot.Literals.String.purs:67]: So. `Char.fromCharCode` is correct per [Prim docs](https://pursuit.purescript.org/builtins/docs/Prim#t:Char):
> A single character (UTF-16 code unit). The JavaScript representation is a normal String, which is guaranteed to contain one code unit. This means that astral plane characters (i.e. those with code point values greater than 0xFFFF) cannot be represented as Char values.

The issue here is that we're parsing something beyond that. In actuality, we need to use CodePoints to accurately parse this. Per [CodePoint docs](https://pursuit.purescript.org/packages/purescript-strings/6.0.1/docs/Data.String.CodePoints#t:CodePoint):
> CodePoint is an Int bounded between 0 and 0x10FFFF, corresponding to Unicode code points.

The fix is to use `(toEnum :: Int -> Maybe CodePoint)`. Proof via Try PS: https://try.purescript.org/?code=FAWw9gJgrgNgpgAgLIEMCWA7BB3AFnAJzmGDRAAcwCAXBABSJigmNIqtoBEVqUA6VAE8ARogAUQ0WL58AlABoEAMwJgQkuIpVqAUlADO1RSBQi4stpRoIAokqVwAxrTF2Hzi2Su03T6nwBhGBR9fUCwDH0weAQxGDAAc092awAVAkE6KCIAZUcCNHJaEIR0zOy4PIKiyw4Ebl5A3BQCBBKA5oJa6wb%2BAEkMYv0EAepurh5%2BHOoCjATwljowTGphkpyAunH6yb4bDCgQWOowfcOLUHQsAC5r23s-BABVDDQxk0wEAF5SjKzc-KFfxEDAsVpfAA8EN%2B5QB1X82DeuACESiMQgYGACAQ8QSCAAJG1yOQ4KCEAAiACMAAYAGL02ltYYsRxkFAwO7YrncrnkgkIfS4MDYfmjPjaEDTWYJACCwzF%2BAAHigWWyYBSafZ7OSsTjEvyUMTSRAKR0WnwTmaCCiWMcwOQELcEFbZHc%2BYTBcL%2BXpDPyrRawFabYgTuRdbiDUayeTOGgHiDHHBOTyU6mU%2B6BUKRYTKbmACy5ykIAC0CAAbABWCsAZgr4f1hMNJOjT30mDxGzobv5nuzCBMDs7fDbc3gJywYhOZyOhYLucddw0CCHwaWKws2KAA

```
main :: Effect Unit
main = TryPureScript.render =<< TryPureScript.withConsole do
  log $ append "10FFFF as decimal:             " $ show $ Int.fromStringAs Int.hexadecimal "10ffff"
  log $ append "Char.toCharCode (top :: Char): " $ show $ Just $ Char.toCharCode top
  log $ append "Difference:                          " $ show $ 1114111 - 65535
  log $ append "Using SCP: " $ show $ map SCP.singleton (toEnum 1114111 :: Maybe SCP.CodePoint)
```

Comment (turlando) [reply test-snapshots/src/snapshots-input/Snapshot.Literals.String.purs:67]: I've read the lexer code a bit. If I understanc correctly the current version of parseEscape which eventually calls parseHexEscape is used by both parseStringEscape (which is called by parseStringPart and parseString) and by parseChar, but only works for the latter. parseStringEscape should use a function that uses CodePoint instead of Char. Right?

Comment (natefaubion) [reply test-snapshots/src/snapshots-input/Snapshot.Literals.String.purs:67]: So the issue here is that the compiler has two different constraints for character escapes:
* When used in a Char literal there is an upper bound of 0xFFFF
* When used in a String literal there is an upper bound of 0x10FFF

And `cst-parser` is using the 0xFFFF upper bound for both.

Comment (natefaubion) [reply test-snapshots/src/snapshots-input/Snapshot.Literals.String.purs:67]: https://github.com/natefaubion/purescript-language-cst-parser/issues/48

**Comment by turlando** [comment]
Upstream has been fixed in [v0.12.3](https://github.com/natefaubion/purescript-language-cst-parser/releases/tag/v0.12.3). I think the PR is ready to be reviewed. Using a regex to extract the hex is kind of brutal but I don't think we have alternatives.

File: src/PureScript/Backend/Chez/Syntax.purs [hunk 2]
```diff
@@ -345,6 +349,25 @@ toChezIdent i (Level l) = case i of
   Just (Ident i') -> i' <> show l
   Nothing -> "_" <> show l
 
+jsonToChezString :: Prim.String -> Prim.String
+jsonToChezString s = unicodeReplace s
+  where
+  unicodeRegex :: R.Regex
+  unicodeRegex = case R.regex """\\u([A-F\d]{4})""" R.Flags.global of
+    Left error -> unsafeCrashWith error
+    Right r -> r
```
purefunctor, line 358: You can use `unsafeRegex` for this, I believe
```suggestion
  unicodeRegex :: R.Regex
  unicodeRegex = R.Unsafe.unsafeRegex """\\u([A-F\d]{4})""" R.Flags.global
```

JordanMartinez (approved):

Issue #91, opened 2023-03-28
Escape identifiers


purefunctor:
Make sure to escape characters that are valid in PureScript grammar e.g. `'` and are not in Chez Scheme.

natefaubion:
Note that "identifiers" in CoreFn are not grammatically PureScript identifiers. They are arbitrary strings.

turlando:
From [TSPL Section 1.1](https://scheme.com/tspl4/intro.html#./intro:h1):

> Scheme programs are made up of keywords, variables, structured forms, constant data (numbers, characters, strings, quoted vectors, quoted lists, quoted symbols, etc.), whitespace, and comments.
>
> Keywords, variables, and symbols are collectively called identifiers. Identifiers may be formed from letters, digits, and certain special characters, including ?, !, ., +, -, *, /, <, =, >, :, $, %, ^, &, _, ~, and @, as well as a set of additional Unicode characters. Identifiers cannot start with an at sign ( @ ) and normally cannot start with any character that can start a number, i.e., a digit, plus sign ( + ), minus sign ( - ), or decimal point ( . ). Exceptions are +, -, and ..., which are valid identifiers, and any identifier starting with ->. For example, hi, Hello, n, x, x3, x+2, and ?$&*!!! are all identifiers. Identifiers are delimited by whitespace, comments, parentheses, brackets, string (double) quotes ( " ), and hash marks( # ). A delimiter or any other Unicode character may be included anywhere within the name of an identifier as an escape of the form \xsv;, where sv is the scalar value of the character in hexadecimal notation.

From [CSUG, Section 1.1](https://cisco.github.io/ChezScheme/csug9.5/intro.html#./intro:h1)

> Chez Scheme extends the syntax of identifiers in several ways. First, the sequence of characters making up an identifier's name may start with digits, periods, plus signs, and minus signs as long as the sequence cannot be parsed as a number. For example, 0abc, +++, and .. are all valid identifiers in Chez Scheme. Second, the single-character sequences { and } are identifiers. Third, identifiers containing arbitrary characters may be printed by escaping them with \ or with |. \ is used to escape a single character (except 'x', since \x marks the start of a hex scalar value), whereas | is used to escape the group of characters that follow it up through the matching |. For example, \||\| is an identifier with a two-character name consisting of the character | followed by the character \, and |hit me!| is an identifier whose name contains a space.

Even if ' is not explicitly mentioned in the above paragraphs I know it's legal to use it as a trailing character in scheme symbols.

```
[...]

PR #109, opened 2023-04-02
Add conversion for uncurried effect functions

purefunctor:


purefunctor:
I was gonna add `assert` testing for this but I ran into #91 for that specific package.

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -137,28 +137,26 @@ codegenExpr codegenEnv@{ currentModule } s = case unwrap s of
...
+  UncurriedApp f p ->
+    S.chezUncurriedApplication (codegenExpr codegenEnv f) (codegenExpr codegenEnv <$> p)
+  UncurriedAbs a e ->
+    S.chezUncurriedFunction (uncurry S.toChezIdent <$> a) (codegenExpr codegenEnv e)
+  UncurriedEffectApp f p ->
+    S.chezThunk $ S.chezUncurriedApplication (codegenExpr codegenEnv f)
+      (codegenExpr codegenEnv <$> p)
+  UncurriedEffectAbs a e ->
+    S.chezUncurriedFunction (uncurry S.toChezIdent <$> a)
+      (S.chezUnthunk $ codegenPureChain codegenEnv e)
```
JordanMartinez, line 154: Shouldn't these not be thunked?

purefunctor: For `UncurriedEffectApp`, applying `f` to `p` runs the effect immediately without the `(lambda () ..)` thunking. Likewise for `UncurriedEffectAbs`, since `e` evaluates to a thunked value, it has to be unthunked.

_I think_ this can be replaced this way though, but I'm not entirely sure...
```suggestion
  UncurriedEffectApp f p ->
    S.chezThunk $ S.chezUncurriedApplication (codegenExpr codegenEnv f)
      (codegenExpr codegenEnv <$> p)
  UncurriedEffectAbs a e ->
    S.chezUncurriedFunction (uncurry S.toChezIdent <$> a)
      (codegenChain effectBlockMode codegenEnv e)
      -- using codegenEffectChain adds thunking here, which we don't want to do...
```

natefaubion: Representationally, EffectFn should always be identicaly to an uncurried function. That is, it doesn't have any additional `(..)` invocation as an `Effect` does. For example, EffectFn2 it is not equivalent to `Fn2 ... (Effect a)`, it is equivalent to `Fn2 ... a`, just impure to the type system. `runEffectFn` function however, _do_ return an `Effect`. The reason for this is that, within an effect _block_ (that is, under an EffectBind), EffectFn functions do not have additional `()` overhead, so multiple calls to EffectFns can be invoked without overhead _in addition to_ the thunk created for the given block.

For example, in `backend-es`:

https://github.com/aristanetworks/purescript-backend-optimizer/blob/85a70a4376bd6b5ab7ae13a6ee1afc8974e06f9c/backend-es/src/PureScript/Backend/Optimizer/Codegen/EcmaScript/Convert.purs#L269-L278

https://github.com/aristanetworks/purescript-backend-optimizer/blob/85a70a4376bd6b5ab7ae13a6ee1afc8974e06f9c/backend-es/src/PureScript/Backend/Optimizer/Codegen/EcmaScript/Convert.purs#L396

https://github.com/aristanetworks/purescript-backend-optimizer/blob/blob/85a70a4376bd6b5ab7ae13a6ee1afc8974e06f9c/backend-es/src/PureScript/Backend/Optimizer/Codegen/EcmaScript/Convert.purs#L475-L476

purefunctor:
Blocked by #116 again. Since I discovered the bug while working on this PR, I'll just add it here as well 😅

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -137,25 +137,21 @@ codegenExpr codegenEnv@{ currentModule } s = case unwrap s of
     S.Identifier $ coerce $ S.toChezIdent i l
   Lit l ->
     codegenLiteral codegenEnv l
+
```
JordanMartinez, line 140: Was this inserted line intended?

test/Main.purs
```diff
@@ -159,15 +159,17 @@ runSnapshotTests { accept, filter } = do
               , moduleName: name
               }
             case result of
-              Left err | matchesFail err.message failsWith -> do
+              Left { message } | matchesFail message failsWith -> do
+                pure true
+              Left { message } -> do
```
JordanMartinez, line 164: Should this become
```purs
Left { message }
  | matchesFail ... ->
  | otherwise ->
```

test/Main.purs
```diff
@@ -159,15 +159,17 @@ runSnapshotTests { accept, filter } = do
...
                 Console.log $ withGraphics (foreground Red) "✗" <> " " <> name <> " failed."
-                Console.log err.message
+                Console.log message
                 pure false
               Right _ | isJust failsWith -> do
                 Console.log $ withGraphics (foreground Red) "✗" <> " " <> name <>
                   " succeeded when it should have failed."
                 pure false
-              _ ->
+              Right _ -> do
```
JordanMartinez, line 172: Similarly here?
```purs
Right _
  | isJust failsWith ->
  | otherwise -> 
```

JordanMartinez:
Otherwise, the code looks good.

JordanMartinez (approved):

PR #111, opened 2023-04-03
Escape identifiers before printing

### LINKED ISSUE #91: Escape identifiers

JordanMartinez:
Fixes #91 by only escaping identifiers before we print them. Much cleaner than #110.

### DIFF
--- a/src/PureScript/Backend/Chez/Syntax.purs
+++ b/src/PureScript/Backend/Chez/Syntax.purs
@@ -10,9 +10,11 @@ import Data.Array.NonEmpty as NEA
 import Data.Array.NonEmpty as NonEmptyArray
 import Data.Maybe (Maybe(..), maybe)
 import Data.Newtype (class Newtype, un)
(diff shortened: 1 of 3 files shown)


purefunctor:
I tried a similar approach to #110 where I newtyped occurrences of `String` into `ChezIdentifier` but I immediately found it an ergonomic nightmare (at least while refactoring) since we're often constructing/deconstructing these types.

I think this is a reasonable approach—exhaustive pattern matching also helps catch when new variants are added. In the future, we could also probably reorganize this once more `ChezLibrary -> ChezLibrary` functions come in.

purefunctor (approved):

JordanMartinez:
Could you submit a PR for just the accessor prop? I can't merge this because CI doesn't pass.

purefunctor:
@JordanMartinez Done! #112

PR #112, opened 2023-04-04
Implement record accessor

purefunctor:


JordanMartinez:
Can you add a test? Maybe ‘test1 = _.foo’

JordanMartinez (approved):

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -157,8 +157,10 @@ codegenExpr codegenEnv@{ currentModule } s = case unwrap s of
   UncurriedEffectAbs _ _ ->
     S.Identifier "uncurried-effect-ab"
 
-  Accessor _ (GetProp _) ->
-    S.Identifier $ "accessor-get-prop"
+  Accessor e (GetProp i) ->
+    S.chezUncurriedApplication
+      (S.Identifier $ scmPrefixed "hashtable-ref")
+      [ codegenExpr codegenEnv e, S.String $ Json.stringify $ Json.fromString i, S.Boolean false ]
```
JordanMartinez, line 163: On second thought, shouldn't this throw an error? The `false` here is the default value returned when the record doesn't contain the property.

PR #114, opened 2023-04-04
Escape chars in codegen

### LINKED ISSUE #65: Fix Char and String literals emission

JordanMartinez:
Fixes #65 (the character part)

I decided to only print chars between 21 and 127 because that's what I could input into the Scheme repl without it complaining. This restriction might be incorrect, but this at least unblocks us.

### DIFF
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -240,7 +240,7 @@ codegenChar :: Char -> ChezExpr
 codegenChar c = S.Char $ append """#\""" $ escapeChar $ toCharCode c
   where
   escapeChar code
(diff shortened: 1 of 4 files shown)


turlando (approved):
Thanks for the extra documentation!

JordanMartinez:
Don't merge this until after #109 gets merged just to reduce potential merge conflicts.

PR #115, opened 2023-04-04
Add flake.nix

purefunctor:


f-f:
Is there a reason for moving to flakes? I am using `lorri`, which does not work with flakes yet, but works with a traditional nix-shell.

purefunctor:
If anything, I just like `nix develop` over `nix-shell` 😄—I haven't used Nix in a while so I wasn't aware that lorri doesn't work as consistently with flakes.

f-f:
I personally prefer flakes too, but this lorri issue has prevented me from switching more repos.
I'm having trouble with it anyways lately, let me see if I can switch to something else

purefunctor:
I'm currently using `nix-direnv` for my WSL setup and it works decently well with Emacs and VS Code. I've yet to try it on flakes though but I think it supports it concretely?

f-f (approved):

PR #117, opened 2023-04-05
Fix record accessor defalt value usage

### LINKED ISSUE #113: Update `AccessorProp` to throw error for default value
See for context https://github.com/purescm/purescm/pull/112/files#r1157034191

[hashtable-ref](https://scheme.com/tspl4/objects.html#./objects:s281)

turlando:
Fixed #113.

Just a proof of concept for now.

### DIFF
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -13,7 +13,7 @@ import Data.String.CodeUnits as CodeUnits
 import Data.Tuple (Tuple(..), uncurry)
 import Data.Tuple as Tuple
 import Partial.Unsafe (unsafeCrashWith)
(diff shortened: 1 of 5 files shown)


turlando:
Using a procedure we basically lose context of where the exception originated. The easiest workaround is to turn `hashtable-get` into a macro, but generating relatively complex Schmee code from PS is *really* hard right now.

JordanMartinez:
So, this code basically works by doing the following:
1. defines a lazy symbol via `gensym` that is unique across the entire program
2. defines a wrapper around `scm:hashtable-re` that returns this unique lazy symbol as the default value
3. if the returned value is that symbol, throw an error. Otherwise, return the value.

The only con to this that you've found is that the error won't be local to the actual `hashtable-ref` usage. Instead, it'll be local to the runtime module. The workaround to that problem is to define this as a macro, but `ChexExpr` doesn't currently support that.

JordanMartinez:
Given that we're going with the SRFI approach, I believe this PR can be closed now.

PR #118, opened 2023-04-05
Fix code generation for effect chain

### LINKED ISSUE #116: Final expression in codegenChain not being run
This has to do with how codegenChain is currently defined. Reading through initially, this might have to do with how EffectPure is being unwrapped as a special case rather than being handled in the "final" step in the accumulation.

purefunctor:
Closes #116. Should finally unblock #109

### DIFF
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -227,42 +227,50 @@ codegenLiteral codegenEnv = case _ of
   LitArray a -> S.vector $ codegenExpr codegenEnv <$> a
   LitRecord r -> S.record $ (map $ codegenExpr codegenEnv) <$> r
 
-type BlockMode = { effect :: Boolean }
+type ChainMode = { effect :: Boolean }
 
-pureBlockMode :: BlockMode
-pureBlockMode = { effect: false }
+pureChainMode :: ChainMode
+pureChainMode = { effect: false }
 
-effectBlockMode :: BlockMode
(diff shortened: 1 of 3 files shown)


src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -227,43 +227,51 @@ codegenLiteral codegenEnv = case _ of
...
         ]
 
+  collect :: Array _ -> NeutralExpr -> ChezExpr
+  collect bindings expression = case unwrap expression of
+    Let i l v e' ->
+      collect (Array.cons { i, l, v: codegenExpr codegenEnv v } bindings) e'
+    EffectPure e' | chainMode.effect ->
+      finish false bindings e'
+    EffectBind i l v e' | chainMode.effect ->
+      collect (Array.cons { i, l, v: S.chezUnthunk $ codegenExpr codegenEnv v } bindings) e'
```
JordanMartinez, line 269: Rather than consing here and reversing in `finish`, why not `snoc` so that reversing isn't needed?

purefunctor: Ah, `snoc` is technically cheaper than `cons` for arrays, right?

purefunctor: or well, I don't think it'd matter much in this specific instance. `snoc` is indeed much cleaner

JordanMartinez (approved):
I'm still not sure if this is the best way to codegen effect but I'm approving this so we can unblocked your other PR.

PR #119, opened 2023-04-06
Refactor ChezExpr to support pretty-printing

### LINKED ISSUE #95: More structured data type for Scheme expressions
Right now we're emitting raw forms (lists and atoms). In order to have better pretty-printing and, most of all, cleaner code it might be a good idea to produce a richer data type representing Scheme code.

@JordanMartinez and @purefunctor pointed out that we should keep a way to emit "raw" scheme forms using the data type we already have (ChezExpr), something like

```
data ChezExpr
  = ManualExpr ChezManualExpr
  | KnownExpr ChezKnownExpr
```

I'm opening this issue to discuss the actual shape of such data structure and the constructors we will need.

JordanMartinez:
Fixes #95.

- Drops old unused printers
- Better distinguishes between syntax and printers
- Pretty-prints `cond`, `let`, and `lambda`. Everything else I wasn't sure should be touched or isn't helped
- Drops the `Prim.` module alias
- Renames function abstraction/application to `mkFn1`/`runFn1`-like terminology.


JordanMartinez:
This is ready for review.

test-snapshots/src/snapshots-output/Snapshot.Let.ss
```diff
@@ -10,4 +10,8 @@
 
   (scm:define letChain
     (scm:lambda (x0)
-      (scm:let* ((a1 (scm:fx+ x0 x0)) (b2 (scm:fx+ a1 a1)) (c3 (scm:fx+ b2 b2))) (scm:fx+ (scm:fx+ (scm:fx+ a1 b2) c3) (scm:fx* c3 c3))))))
+      (scm:let*
+        ([a1 (scm:fx+ x0 x0)]
+        [b2 (scm:fx+ a1 a1)]
+        [c3 (scm:fx+ b2 b2)])
```
purefunctor, line 16: These bindings should be aligned like this, right?
```suggestion
        ([a1 (scm:fx+ x0 x0)]
         [b2 (scm:fx+ a1 a1)]
         [c3 (scm:fx+ b2 b2)])
```

JordanMartinez: Done.

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -83,6 +86,34 @@ codegenModule { name, bindings, imports, foreign: foreign_ } =
...
+  DefineRecordType i fields ->
+    [ recordTypeUncurriedConstructor i
+    , recordTypePredicate i
+    ] <> map (recordTypeAccessor i) fields
+
+resolve :: ModuleName -> Qualified Ident -> String
+resolve _ (Qualified Nothing (Ident i)) = i
+resolve currentModule (Qualified (Just m) (Ident i))
+  | currentModule == m = i
+  | otherwise = coerce m <> "." <> i
```
purefunctor, line 108: This is entirely optional, but is there a good alternative name for `resolve`? I was thinking `flattenQualified`.

purefunctor (approved):
LGTM, the suggestion here is optional

src/PureScript/Backend/Chez/Convert.purs
JordanMartinez: I think `flattenQualified` makes sense.

PR #120, opened 2023-04-08
Convert LetRec

JordanMartinez:


test-snapshots/src/snapshots-input/Snapshot.Let.purs
```diff
@@ -11,3 +11,18 @@ letChain x =
...
+
+isEven :: Int -> Boolean
+isEven x
+  | x == 0 = true
+  | otherwise = not $ isOdd (x - 1)
+
+isOdd :: Int -> Boolean
+isOdd x
+  | x == 1 = true
+  | otherwise = not $ isEven (x - 1)
```
purefunctor, line 28: I think this can be defined as:
```suggestion
isEven :: Int -> Boolean
isEven x
  | x == 0 = true
  | otherwise = isOdd (x - 1)

isOdd :: Int -> Boolean
isOdd x
  | x == 0 = false
  | otherwise = isEven (x - 1)
```

We can use this to check proper treatment of tail calls.
TIL: Scheme has proper treatment of tail calls and as such, supports mutual tail recursion:
```scheme
> (define is-odd?
    (trace-lambda is-odd? (x)
      (if (= x 0) #f (is-even? (- x 1)))))
> (define is-even?
    (trace-lambda is-even? (x)
      (if (= x 0) #t (is-odd? (- x 1)))))
> (is-odd? 10)
|(is-odd? 10)
|(is-even? 9)
|(is-odd? 8)
|(is-even? 7)
|(is-odd? 6)
|(is-even? 5)
|(is-odd? 4)
|(is-even? 3)
|(is-odd? 2)
|(is-even? 1)
|(is-odd? 0)
|#f
#f
> (is-even? 10)
|(is-even? 10)
|(is-odd? 9)
|(is-even? 8)
|(is-odd? 7)
|(is-even? 6)
|(is-odd? 5)
|(is-even? 4)
|(is-odd? 3)
|(is-even? 2)
|(is-odd? 1)
|(is-even? 0)
|#t
#t
```

JordanMartinez: Fixed.

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -89,7 +89,7 @@ codegenModule { name, bindings, imports, foreign: foreign_ } =
 
 definitionIdentifiers :: ChezDefinition -> Array String
 definitionIdentifiers = case _ of
-  DefineValue i _ -> [ i ]
+  Define i _ -> [ i ]
```
JordanMartinez, line 92: Reason for this change. If #117 adds macros to `ChezDefinition` via `define-syntax`, then we should make the constructors match the identifiers they represent.

JordanMartinez:
CI builds. Can this get an approval?

purefunctor (approved):

PR #126, opened 2023-04-10
Use if for branches with single conditions

purefunctor:
Use `if` instead of `cond` when there's only a single condition for a `Branch`


purefunctor:
This is also waiting for https://github.com/aristanetworks/purescript-backend-optimizer/pull/49

f-f:
@purefunctor is it worth bringing this up to date, or should we close it?

PR #128, opened 2023-04-11
Convert PrimUndefined

purefunctor:
This converts `PrimUndefined` to a constant symbol `'undefined`. The other choice was `(void)` but using a constant allows us to know that it came from PS code (or potentially Chez Scheme FFI)


turlando (approved):

JordanMartinez:
I'm requesting that we use `gensym "ps-undefined"` (or some name like that) to prevent this symbol from being equal to another one somewhere in the program.

JordanMartinez (approved):
Code otherwise looks good to me.

Issue #129, opened 2023-04-11
Convenience macros for hashtable operations


purefunctor:
Right now, there's a bit of syntactic overhead when creating a hashtable for records:
```lisp
  (scm:define main
    (scm:let*
      ([r0 (scm:letrec* (($record (scm:make-hashtable scm:string-hash scm:string=?))) (scm:hashtable-set! $record "fooBarBaz" 5) $record)]
       [s1 (recordUpdate r0)]
       [_2 (Test.Assert.assert (scm:fx=? (recordAccess r0) 5))])
        (scm:lambda ()
          (scm:let ([_ (_2)])
            ((Test.Assert.assert (scm:fx=? (recordAccess s1) 10))))))))
```

We could define a macro in the runtime module that allows us to do:
```lisp
(rt:record
  ["fooBarBaz" 5]
  ..)
```

Likewise, record updates could also be prettier:
```lisP
(rt:record-update $record
  ["fooBarBaz" 10]
  ..)
```

Thoughts?

purefunctor:
SRFI 125 (https://srfi.schemers.org/srfi-125/srfi-125.html) defines a convenience constructor called `hash-table` with the following form:
```scheme
 (hash-table comparator [ key value ] ...)
```
I reckon if we go through with #131 we wouldn't really need to define our own macro in our runtime.

PR #130, opened 2023-04-11
Convert Update

purefunctor:
This converts `Update` to a form that:
1. Copies the hashtable
2. Sets the properties

In #129 I mentioned we could do pretty-printing for this and I could work on that in the future (tomorrow).

### DIFF
--- a/src/PureScript/Backend/Chez/Syntax.purs
+++ b/src/PureScript/Backend/Chez/Syntax.purs
@@ -156,8 +156,8 @@ eqQ x y = runUncurriedFn (Identifier $ scmPrefixed "eq?") [ x, y ]
 vector :: Array ChezExpr -> ChezExpr
 vector = List <<< Array.cons (Identifier $ scmPrefixed "vector")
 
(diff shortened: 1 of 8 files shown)


src/PureScript/Backend/Chez/Syntax.purs
```diff
@@ -156,6 +156,9 @@ eqQ x y = runUncurriedFn (Identifier $ scmPrefixed "eq?") [ x, y ]
 vector :: Array ChezExpr -> ChezExpr
 vector = List <<< Array.cons (Identifier $ scmPrefixed "vector")
 
+hashtableCopy :: ChezExpr -> ChezExpr
+hashtableCopy h = runUncurriedFn (Identifier $ scmPrefixed "hashtable-copy") [ h ]
+
```
JordanMartinez, line 161: Per [the `hashtable-copy` docs](https://scheme.com/tspl4/objects.html#./objects:s286), this form produces an immutable hashtable. So, does setting the copy's field even work?

PR #132, opened 2023-04-12
Major and minor refactorings

purefunctor:
We should probably do a merge commit for this to preserve history?

### DIFF
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -2,7 +2,6 @@ module PureScript.Backend.Chez.Convert where
 
 import Prelude
 
-import Data.Argonaut as Json
 import Data.Array as Array
 import Data.Array.NonEmpty as NEA
 import Data.Array.NonEmpty as NonEmptyArray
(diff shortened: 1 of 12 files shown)


JordanMartinez:
> * Move common forms to Common.purs

Why make the above change in your first commit? I don't understand why this should be done and without knowing that, it just seems like a pointless change to me.

purefunctor:
> > * Move common forms to Common.purs
> 
> Why make the above change in your first commit? I don't understand why this should be done and without knowing that, it just seems like a pointless change to me.

I was thinking of also moving the `scmPrefixed` values into the `Constants` file, where they would also be wrapped in `Identifier`. The dependency tree would look like:
```
Syntax -> Constants -> Common
Syntax -> Common
```

f-f:
@purefunctor if these refactorings are unrelated it might be worth to have each be in their own pull request, so they can have separate discussions

f-f:
@purefunctor is it worth to bring this up to date, or should we close it?

purefunctor:
I think this should be closed, I'll re-open PRs over time for the changes here.

PR #133, opened 2023-04-18
Add initial chez-srfi install script

purefunctor:
Note that the vendoring process also adds compiled Scheme files which may be platform dependent. Right now, I just have it such that the vendor directory isn't tracked through Git though it may be desirable to do so.

When testing, the `vendor` directory can be added into the libdirs while executing `scheme` or copied over to the `output` directory. Likewise, the copying should be what's done for when `purescm` becomes an executable backend.

`purescm`-based projects in the future may make use of a package manager like Akku, and we should document that `chez-srfi` is already being vendored by the backend itself. Alternatively, maybe we could make it so that `chez-srfi` that we bundle exists under some `version`?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@
 /test-snapshots/generated-docs
 /test-snapshots/test-out/
 /.direnv
+/tools/.spago
+/tools/output
+/vendor/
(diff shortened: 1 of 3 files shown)


purefunctor:
Closing this as the PR seems to be bugged. See #134 instead.

PR #134, opened 2023-04-18
Add initial vendoring for chez-srfi

purefunctor:
Note that the vendoring process also adds compiled Scheme files which may be platform dependent, so I just gitignored the `.so` files.

When testing, the vendor directory can be added into the libdirs while executing scheme or copied over to the output directory. Likewise, the copying should be what's done for when purescm becomes an executable backend.

purescm-based projects in the future may make use of a package manager like Akku, and we should document that chez-srfi is already being vendored by the backend itself. Alternatively, maybe we could make it so that chez-srfi that we bundle exists under some version?


JordanMartinez:
I'm confused. Don't we only need the `hashtables` one? Why include all these other ones?

f-f:
I also think we shouldn't include all these in the repo.
We could stash them in a separate repo and import the ones we need though.

purefunctor:
> I'm confused. Don't we only need the hashtables one? Why include all these other ones?
In the future, there are probably users who would use a package manager like Akku to install extra packages to import in the future. Having the full `chez-srfi` package vendored simplify the need to install the package through Akku if some other package needs it. `chez-srfi` is relatively stable though, so I don't think we'll see any breakage if we just provide `chez-srfi` ourselves rather than provide only some parts of it and then have the user install the more "complete" version if need be.

> I also think we shouldn't include all these in the repo.
Yeah, I think keeping track of the `revision` in the importer tool should be enough.

JordanMartinez:
> In the future, there are probably users who would use a package manager like Akku to install extra packages to import in the future. Having the full chez-srfi package vendored simplify the need to install the package through Akku if some other package needs it.

Seems related to this issue: https://gitlab.com/akkuscm/akku/-/issues/40

JordanMartinez:
Also, the `chez-srfi` seems to impact windows: https://github.com/arcfide/chez-srfi/issues/76

purefunctor:
> Also, the `chez-srfi` seems to impact windows: [arcfide/chez-srfi#76](https://github.com/arcfide/chez-srfi/issues/76)

If that's the case, then perhaps we could extract what we need manually. IIRC, the SRFIs that we need are [125](https://srfi.schemers.org/srfi-125/) for providing records and [214](https://srfi.schemers.org/srfi-214/) for providing mutable arrays.

I guess the challenge here is that if we vendor these SRFIs privately, it'll be completely separate from `chez-srfi` and so any other Scheme package that depends on it would break; though, I'm not entirely sure just how significant this breakage could be.

PR #135, opened 2023-05-08
Vendor SRFI 214 and SRFI 125

purefunctor:
This PR adds vendoring for SRFI 214 (Flexvectors) and SRFI 125 (Intermediate hash tables). These are used to represent the `Array` type and record types respectively. Likewise, these vendored files are not meant to be publicly consumable aside from FFI implementers. Code generation based on these will be added on a future pull request.


f-f (approved):
Sorry for delay, had a very busy week at work.

This looks great, a few questions:
- I assume the other SRFIs included are the ones that 125 and 214 depend on?
- How do we add new ones? Would be useful to have some docs on that. The included ones reference a `install-chezscheme.sls` script which we don't have. Wonder if we should strip that out to not be confusing.

f-f:
> The process involves ..

Could you add this to the readme so it's documented if anyone wants to go at it?

f-f (approved):
Wonderful, thank you!

PR #136, opened 2023-05-17
Fix cache key in CI

purefunctor:
This changes the cache key to depend on spago.yaml files instead of  *.dhall files from old Spago


purefunctor:
Weird, something else is broken other than this

purefunctor:
Got it! I created #137

PR #138, opened 2023-05-27
Integrate SRFIs to runtime

purefunctor:
This PR integrates the added SRFIs to the runtime. Arrays and records are now represented by SRFI 214 and SRFI 125 respectively. This is the second half of #135.


JordanMartinez (approved):
I'm mainly approving this so we can make more progress on things.

PR #140, opened 2023-07-08
Export only explicitly exported identifiers

### LINKED ISSUE #106: Don't export all the library definitions
Due to #97, we're exporting all the ChezDefinitions. We should only export those that are explicitly in the exports set.

anttih:
Only export identifiers set in the exports list. This fixes #106.

We filter the identifiers after code generation which means we need to keep the original purs identifiers in the `ChezDefinition`. The only case where we don't keep track of the original identifier name in the `ChezDefinition` is the zero arity constructor for which we generate a predicate. So I added a new `ChezDefinition` for the predicate and then delay the generation of the predicate identifier name until the `Printer`.

### DIFF
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -46,11 +46,12 @@ codegenModule { name, bindings, exports, imports, foreign: foreign_ } =
 
     exports' :: Array ChezExport
     exports' =
-      let exportedIdentifiers = flip Array.filter definitions case _ of
-            Define i _ -> Set.member (Ident i) exports
-            DefineRecordType i _ -> Set.member (Ident i) exports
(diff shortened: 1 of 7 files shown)


f-f (approved):
This looks great, thanks Antti!

natefaubion:
Since the backend optimizer inlines code, you can’t only export things exported in the PS source, as an inlined definition may refer to an unexpired definition.

anttih:
@natefaubion so do we have to always export everything?

natefaubion:
In general, yes, the expectation is that all symbols which are reachable from PureScript are available in some way across module boundaries, as inlining necessarily breaks high-level module abstractions. The reason why `exports` are still useful is that one might mangle private names differently or otherwise reference them in a different way.

anttih:
Ok, then I guess we can close this and #106.

PR #141, opened 2023-07-08
Fix references to arity length of one constructors

### LINKED ISSUE #139: Call to data constructor of arity 1 generates invalid code
```
data Foo = Foo Int

foo :: Foo
foo = Foo 1
```

Generates

```
(scm:define foo (Foo* 1))
```

The constructor name currently is `Foo`, not `Foo*`.

anttih:
Fixes #139


JordanMartinez:
Just FYI. Any new contributors need to add themselves to CONTRIBUTORS.md.

f-f (approved):

PR #148, opened 2023-08-04
Fix array-ref and array-length references

anttih:
* Fixes array indexing to reference the correct runtime symbol
* Fixes `array-length` reference and adds that to the runtime


f-f (approved):

f-f:
It'd be useful for these kinds of changes to have runtime tests where we check that running the scheme gives us a correct result

PR #149, opened 2023-08-21
Add initial version of CLI tool for compiling and bundling

anttih:
Adds a CLI with two commands:

* `build` - compiles the corefn files to scheme code using the backend-optimizer.
* `bundle-app` - compiles the scheme files to a single "program" file that has no dependencies except the scheme runtime. This file can be run with `scheme --program <output-dir>/main`.

How `bundle-app` works is the following:

* It creates a scheme file that imports the main module and invokes the `main` function. This intermediate file is required because the `compile-program` call in the next step requires the filename as a string.
* It then runs the following "script" by sending it to `scheme`:

```scheme
(top-level-program (import (chezscheme))
  (optimize-level 3)
  (compile-imported-libraries #t)
  (generate-wpo-files #t)
  ;; Compile the "main". This generates the `.so` and `.wpo` files for _all_
  ;; code because we set `compile-imported-libraries` and `generate-wpo-files`.
  (compile-program "output-chez/main.ss")
  ;; Compile all `.wpo` files into a single "output-chez/main" file.
  (compile-whole-program "output-chez/main.wpo" "output-chez/main"))
```

Todo:

* [x] ~~Relies on the runtime module being compiled and placed in the `vendor/` directory.~~ This is not the problem. We can just regenerate the runtime scheme file.
* [x] Don't write `.so` and `.wpo` files for the runtime libs into the source directory, but instead compile them to the output-dir.


anttih:
The build failure will be fixed by https://github.com/purescm/purescm/pull/150

f-f:
@anttih I tried this out on the `test-snapshots` project we have in the repo, and `build` works, but `bundle-app` errors with `Exception: library (Test.Main lib) not found`

Also, why use `output-chez` and not just `output`?

anttih:
> @anttih I tried this out on the `test-snapshots` project we have in the repo, and `build` works, but `bundle-app` errors with `Exception: library (Test.Main lib) not found`

Ah, I wasn't using the arg `--main` at all. It's now being used. We'll need to add the check whether the passed module has a callable `main`. Looks like the test suite has that, I'll try to add that.
 
> Also, why use `output-chez` and not just `output`?

I changed this to just `output/`.

anttih:
I looked into checking if the `main` is really an `Effect Unit`, and I'm not really sure how to do it or whether it's even worth doing. The test suite parses the original source file with the CST parser, but in the CLI tool we don't have the source file necessarily because all we see are the `corefn` files.

Right now if you try to `bundle-app` a module that doesn't have a `main`, the bundling will fail because of the checks that `chez` is doing:

```
Exception: attempt to reference unbound identifier main at line 1, char 27 of output/main.ss
```

it will not however check if `main` has the correct signature. What do you think of just not doing that check for now?

f-f:
> What do you think of just not doing that check for now?

Yeah I think it's fine to not worry about this for now!

f-f (approved):
To validate this I have run:
```
npm run build
```
in the root of the repo, to build the new CLI tool.

Then in the `test-snapshots` folder (which contains a purescm project):
```
spago build --purs-args "-g corefn" # to compile the purescript into corefn
../index.js build # to compile the corefn into scheme
../index.js bundle-app --main Snapshot.Effect # to create the scheme entrypoint file
scheme --program output/main # to run this scheme file
```

It all seems working well - thanks @anttih!

I guess we'll eventually make this compatible with spago backend support, so that one can add the `backend: purescm` config bit to the spago project, and spago will do the calling of these build steps

PR #153, opened 2023-09-15
Add support for directives file

anttih:
Adds the ability to pass `--directives <file>` to the `build` command.

### DIFF
--- a/src/Main.purs
+++ b/src/Main.purs
@@ -6,7 +6,7 @@ import ArgParse.Basic (ArgParser)
 import ArgParse.Basic as ArgParser
 import Data.Array as Array
 import Data.Either (Either(..), isRight)
-import Data.Maybe (fromMaybe)
+import Data.Maybe (Maybe, fromMaybe)
 import Data.Monoid (power)
 import Data.Newtype (unwrap)
 import Data.Set as Set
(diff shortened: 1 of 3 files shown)


src/Main.purs
```diff
@@ -112,7 +118,11 @@ main cliRoot = do
     Left err ->
       Console.error $ ArgParser.printArgError err
     Right (Build args) ->
-      launchAff_ $ runBuild args
+      flip runAff_ (runBuild args) case _ of
+        Right _ -> pure unit
+        Left err -> do
+          Console.error (message err)
+          Process.exit 1
```
anttih, line 125: Any nicer way to write this?

f-f: I think this is the boring but readable way to write this 😄

f-f (approved):

src/Main.purs
JordanMartinez: Technically, we should be using `Process.setExitCode 1` because `Process.exit 1` could cause content not to be flushed to `stdout`/`stderr` before the Node process exits. See https://nodejs.org/dist/latest-v18.x/docs/api/process.html#processexitcode

anttih: Our version of `node-process` doesn't have that function. I'll fix all these in a separate PR. Thanks!

PR #154, opened 2023-09-15
Change record codegen to use alist->hash-table

anttih:
I discovered there's a strange bug with how we construct records. If you write a record literal where the first value is a negative `Int` you get a runtime error from Chez. The generated code is something like this currently:

```
> (hash-table (make-comparator string? string=? string<? string-hash) "foo" -2 "bar" 4)
Exception in make-hashtable: invalid size argument -2
```

The [SRFI](https://srfi.schemers.org/srfi-125/srfi-125.html#Constructors) states that:

> The meaning of any further arguments is implementation-dependent. However, implementations which support the ability to specify the initial capacity of a hash table should interpret a non-negative exact integer as the specification of that capacity.

which doesn't really explain this behavior, but at least that there is something going on with the "size" here. I'm not able to make `hash-table` work with a negative first value. It doesn't look like we are using it incorrectly, because if you pass in a positive int, it seems to work fine.

Anyway, it appears the `hash-table` function is maybe a bit more useful when using hash tables in an imperative mutable manner, and that for our purposes the `alist->hash-table` seems a better fit. So I've changed our codegen to use `alist->hash-table` instead. Then the generated code is something like this:

```
> (alist->hash-table (list (cons "foo" -2) (cons "bar" 4)) (make-comparator string? string=? string<? string-hash))
#<hashtable>
```

I tried writing a test but I cannot construct a negative `Int` literal because the `negate` function is not available (it's in Prelude).

### DIFF
--- a/src/Main.purs
+++ b/src/Main.purs
@@ -6,7 +6,7 @@ import ArgParse.Basic (ArgParser)
 import ArgParse.Basic as ArgParser
 import Data.Array as Array
 import Data.Either (Either(..), isRight)
-import Data.Maybe (fromMaybe)
+import Data.Maybe (Maybe, fromMaybe)
 import Data.Monoid (power)
 import Data.Newtype (unwrap)
 import Data.Set as Set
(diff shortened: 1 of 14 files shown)


f-f:
I think we should have a test for this, and I think it'd be possible to use FFI to get a negative into?

anttih:
> I think we should have a test for this, and I think it'd be possible to use FFI to get a negative into?

Duh, yes of course! 

I've added a test, but unfortunately the test fails silently even before getting to evaluate the `make-object` call with `Exception: Record.Unsafe:unsafeSet not implemented.` We have a bunch of asserts that never evaluate because of the missing prelude.

f-f:
> Exception: Record.Unsafe:unsafeSet not implemented

I guess we should port the Prelude things first then?

anttih:
Yeah, I discovered this exact bug while working on the prelude :) This is a bit of a chicken-and-egg problem. One approach could be that I start introducing smaller changes to prelude, but then again we can't run the prelude test suite yet.

f-f:
Ah I see. I guess we could open an issue to track "add test for this" and then merge this

anttih:
This is now actually running the test.

f-f (approved):

PR #155, opened 2023-09-16
Upgrade to package set 41.2.0

f-f:
@JordanMartinez there are a lot of breaking changes in the Node libraries and I am not quite sure the changes I did to port to the new API make sense, so I'd appreciate if you could have a look

### DIFF
--- a/src/Main.purs
+++ b/src/Main.purs
@@ -6,7 +6,7 @@ import ArgParse.Basic (ArgParser)
 import ArgParse.Basic as ArgParser
 import Data.Array as Array
 import Data.Either (Either(..), isRight)
-import Data.Maybe (Maybe, fromMaybe)
+import Data.Maybe (Maybe(..), fromMaybe)
 import Data.Monoid (power)
 import Data.Newtype (unwrap)
 import Data.Set as Set
@@ -21,12 +21,13 @@ import Effect.Class (liftEffect)
 import Effect.Class.Console as Console
 import Effect.Exception (message)
 import Node.Encoding (Encoding(..))
+import Node.EventEmitter (on_)
 import Node.FS.Aff as FS
 import Node.FS.Perms (mkPerms)
 import Node.FS.Perms as Perms
 import Node.FS.Stats as Stats
 import Node.FS.Stream (createReadStream, createWriteStream)
-import Node.Library.Execa (ExecaError, ExecaSuccess, execa)
+import Node.Library.Execa (ExecaResult, execa)
 import Node.Library.Execa.Which (defaultWhichOptions, which)
 import Node.Path (FilePath)
 import Node.Path as Path
@@ -122,7 +123,7 @@ main cliRoot = do
         Right _ -> pure unit
         Left err -> do
           Console.error (message err)
-          Process.exit 1
+          Process.exit' 1
     Right (Bundle arg) ->
       launchAff_ $ runBundle cliRoot arg
 
@@ -195,18 +196,18 @@ runBundle cliRoot args = do
     , "      (compile-whole-program \"" <> mainWpoPath <> "\" \"" <> outPath <> "\")"
     , ")))"
     ]
-  case res of
-    Left err -> liftEffect $ Process.exit (fromMaybe 1 err.exitCode)
-    Right _ -> Console.log $ "Created " <> outPath
+  case res.exitCode of
+    Just 0 -> Console.log $ "Created " <> outPath
+    _ -> liftEffect $ Process.exit' (fromMaybe 1 res.exitCode)
 
-evalScheme :: Array String -> String -> Aff (Either ExecaError ExecaSuccess)
+evalScheme :: Array String -> String -> Aff ExecaResult
 evalScheme arguments code = do
   schemeBin <- getSchemeBinary
   spawned <- execa schemeBin arguments identity
   spawned.stdin.writeUtf8End code
   void $ liftEffect $ Stream.pipe spawned.stdout.stream Process.stdout
   void $ liftEffect $ Stream.pipe spawned.stderr.stream Process.stderr
-  spawned.result
+  spawned.getResult
 
 getSchemeBinary :: Aff String
 getSchemeBinary = do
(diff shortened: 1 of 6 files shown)


anttih (approved):

src/Main.purs
```diff
@@ -195,18 +196,18 @@ runBundle cliRoot args = do
     , "      (compile-whole-program \"" <> mainWpoPath <> "\" \"" <> outPath <> "\")"
     , ")))"
     ]
-  case res of
-    Left err -> liftEffect $ Process.exit (fromMaybe 1 err.exitCode)
-    Right _ -> Console.log $ "Created " <> outPath
+  case res.exitCode of
+    Just 0 -> Console.log $ "Created " <> outPath
+    _ -> liftEffect $ Process.exit' (fromMaybe 1 res.exitCode)
```
JordanMartinez, line 201: Using `res.exit` will get you an Either-like type containing either the exit code or the kill signal used to kill the child process.

src/Main.purs
```diff
@@ -231,13 +232,9 @@ copyFile from to = do
...
     src <- createReadStream from
     dst <- createWriteStream to
-    res <- Stream.pipe src dst
-    Stream.onError src (k <<< Left)
-    Stream.onError dst (k <<< Left)
-    Stream.onError res (k <<< Left)
-    Stream.onFinish res (k (Right unit))
+    Stream.pipe src dst
+    src # on_ Stream.errorH (k <<< Left)
+    dst # on_ Stream.errorH (k <<< Left)
```
JordanMartinez, line 237: I think I would move the error stuff above the `pipe` line.

Otherwise, Node docs state:
> One important caveat is that if the Readable stream emits an error during processing, the Writable destination is not closed automatically. If an error occurs, it will be necessary to manually close each stream in order to prevent memory leaks.

So, the `src` errorH needs to call close on `dst` before calling `k <<< Left`.

test/Utils.purs
```diff
@@ -92,13 +94,10 @@ copyFile from to = do
...
     src <- createReadStream from
     dst <- createWriteStream to
-    res <- Stream.pipe src dst
-    Stream.onError src (k <<< Left)
-    Stream.onError dst (k <<< Left)
-    Stream.onError res (k <<< Left)
-    Stream.onFinish res (k (Right unit))
+    Stream.pipe src dst
+    src # on_ Stream.errorH (k <<< Left)
+    dst # on_ Stream.errorH (k <<< Left)
```
JordanMartinez, line 99: Similar situation here.

test/Utils.purs
```diff
@@ -36,22 +39,21 @@ import PureScript.CST.Types as CSTT
...
 execWithStdin :: String -> String -> Aff ExecResult
 execWithStdin command input = makeAff \k -> do
-  childProc <- ChildProcess.exec command defaultExecOptions (k <<< pure)
-  _ <- Stream.writeString (ChildProcess.stdin childProc) UTF8 input mempty
-  Stream.end (ChildProcess.stdin childProc) mempty
-  pure $ effectCanceler $ ChildProcess.kill SIGABRT childProc
+  childProc <- ChildProcess.exec' command identity (k <<< pure)
+  _ <- Stream.writeString' (ChildProcess.stdin childProc) UTF8 input mempty
+  Stream.end' (ChildProcess.stdin childProc) mempty
+  pure $ effectCanceler $ void $ ChildProcess.killSignal SIGABRT childProc
```
JordanMartinez, line 56: Shouldn't this be updated to use `node-execa`?

test/Utils.purs
```diff
@@ -36,22 +39,21 @@ import PureScript.CST.Types as CSTT
...
     Normally code
-      | code > 0 -> Process.exit code
+      | code > 0 -> Process.exit' code
       | otherwise -> k (Right unit)
-    BySignal _ ->
-      Process.exit 1
+    BySignal _ -> Process.exit' 1
   pure $ effectCanceler do
-    ChildProcess.kill SIGABRT childProc
+    void $ ChildProcess.killSignal SIGABRT childProc
```
JordanMartinez, line 49: Same here. Shouldn't this be updated to use `node-execa`?

src/Main.purs
```diff
@@ -195,18 +196,18 @@ runBundle cliRoot args = do
...
+    _ -> liftEffect $ Process.exit' (fromMaybe 1 res.exitCode)
 
-evalScheme :: Array String -> String -> Aff (Either ExecaError ExecaSuccess)
+evalScheme :: Array String -> String -> Aff ExecaResult
 evalScheme arguments code = do
   schemeBin <- getSchemeBinary
   spawned <- execa schemeBin arguments identity
   spawned.stdin.writeUtf8End code
   void $ liftEffect $ Stream.pipe spawned.stdout.stream Process.stdout
   void $ liftEffect $ Stream.pipe spawned.stderr.stream Process.stderr
```
JordanMartinez, line 209: This can be `spawned.stdout.pipeToParentStdout` and `spawned.stderr.pipeToParentStderr`

src/Main.purs
f-f: Does it matter if I just want to know if it exited with 0? Or will it not emit an exit code if there was a kill signal?

src/Main.purs
f-f: How would that look like?

test/Utils.purs
f-f: Some of these changes are over my head and I'd just like to get the patch through for now - we can refactor more in depth in later patches

src/Main.purs
f-f: Great point

src/Main.purs
JordanMartinez: No it does not. `exit` just handles the otherwise type unsafe way of doing this:
```
case res.exitCode, res.signal of
  Just c, _ ->
  _, Just s ->
  _, _ -> unsafeCrashWith "Impossible"
```

src/Main.purs
JordanMartinez:
```purs
src # on_ Stream.errorH \err -> do
  Stream.destroy' dst $ error $ "Got error in src stream: " <> message err
  k $ Left err
dst # on_ Stream.errorH (k <<< Left)
Stream.pipe src dst

test/Utils.purs
f-f: That would be great, thanks!

JordanMartinez (approved):

src/Main.purs
f-f: Thanks!

PR #159, opened 2023-09-23
Update package.json and README

anttih:
* Updates package.json to make it possible to publish an npm package.
* Update README to reflect the correct package name and the fact that this can now actually be installed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,12 +1,17 @@
 {
-  "private": true,
+  "name": "purescm",
+  "version": "1.0.0",
   "main": "index.js",
   "type": "module",
+  "files": [
+    "vendor/**/*.sls",
+    "bundle/index.js"
+  ],
   "bin": {
-    "purs-backend-chez": "./index.js"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,12 +1,17 @@
...
+    "bundle/index.js"
+  ],
   "bin": {
-    "purs-backend-chez": "./index.js"
+    "purescm": "./index.js"
   },
   "scripts": {
-    "test": "spago -x test.dhall test",
+    "test": "spago test",
     "build": "spago build --purs-args '-g corefn,js' && purs-backend-es bundle-module -m Main --int-tags --platform=node --minify --to=./bundle/index.js"
```
f-f, line 15: I believe this would fail with the latest spago, should be changed to this:

```suggestion
    "build": "spago build && purs-backend-es bundle-module -m Main --int-tags --platform=node --minify --to=./bundle/index.js"
```

f-f (approved):

PR #161, opened 2023-09-24
Make sure to compile to corefn in build script

anttih:
`spago` doesn't compile to `corefn` unless told so. We need this for `purs-backend-es` build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,6 @@
   },
   "scripts": {
     "test": "spago test",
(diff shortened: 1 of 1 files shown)


f-f:
The latest Spago will compile to corefn all the time, so we should upgrade to latest instead of doing this. I'm having a look at this (starting from https://github.com/thomashoneyman/purescript-overlay/issues/47)

PR #162, opened 2023-09-24
Update to Spago 0.93.14, switch to purescript-overlay

f-f:
This updates to the latest Spago, which will compile to corefn all the time, and to js only when the backend is not provided.

We also switch from easy-purescript-nix to purescript-overlay for the nix machinery, as it's generally more up to date.

Closes #161

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,21 +1,5 @@
 {
   "nodes": {
-    "easy-purescript-nix": {
-      "flake": false,
-      "locked": {
-        "lastModified": 1680076811,
-        "narHash": "sha256-vfzrVwBntXao3nMi2hkjYlWGUnuyUOVmYi95mQQ0EEY=",
-        "owner": "f-f",
-        "repo": "easy-purescript-nix",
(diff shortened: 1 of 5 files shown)


f-f:
@anttih I had to add a `backend` to the `test-snapshots` project (which I fair I think), but it doesn't seem to be running tests anymore, and I'm not familiar enough with how the test framework works to be able to fix it

f-f:
It looks like the Test machinery is itself compiling the modules, and then comparing them to the snapshots, but I'd prefer that we let the backend hook compile things (so we test that the executable works too) and instead read the files afterwards

anttih:
@f-f yes, the test suite itself performs the role of the purescm backend by compiling the corefn files. It would be a bigger task to change this, and not sure how straight forward that is.

f-f:
I'll have a look

f-f:
@anttih I ported the test setup - the only thing that differs is that we are not checking for runtime failures anymore (the old implementation would look for comments with a `@fails` annotation), but it looks like that was unused so we can reintroduce it later if we need it

anttih (approved):

PR #163, opened 2023-09-28
Publish npm package on push

anttih:


anttih:
This doesn't seem super useful and gets a bit tricky with the versioning.

PR #164, opened 2023-09-29
Add --version flag and scripts to make a prerelease

anttih:
* Adds scripts to bump the version
* Adds a `--version`
* Bump version

I did not publish this release yet, but I will do that before merging.

### DIFF
--- a/spago.yaml
+++ b/spago.yaml
@@ -1,6 +1,11 @@
 package:
   name: purescm
-  version: 1.0.1-pre.0
+  publish:
+    version: 1.0.1
+    license: Apache-2.0
+    location:
+      githubOwner: purescm
+      githubRepo: purescm
   dependencies:
     - aff
     - argonaut
--- a/spago.yaml
+++ b/spago.yaml
@@ -1,5 +1,6 @@
 package:
   name: purescm
+  version: 1.0.1-pre.0
   dependencies:
     - aff
     - argonaut
--- a/src/Version.js
+++ b/src/Version.js
@@ -1,0 +1,1 @@
+export const version = "v1.0.0";
--- a/src/Version.js
+++ b/src/Version.js
@@ -1,1 +1,1 @@
-export const version = "v1.0.0";
+export const version = "v1.0.1-pre.0";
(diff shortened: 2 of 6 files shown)


src/Version.js
```diff
@@ -0,0 +1 @@
+export const version = "v1.0.1-pre.0";
```
f-f, line 1: You can import the generated Spago files to avoid hardcoding this, e.g. [like it's done here](https://github.com/purescript/spago/blob/a4fc6b3cfd3410e8cd0ecb2cf547bda6568aac1c/src/Spago/BuildInfo.purs#L77)

f-f: The file is sitting in `.spago/BuildInfo.purs`

anttih: Wonderful!

spago.yaml
```diff
@@ -1,5 +1,6 @@
 package:
   name: purescm
+  version: 1.0.1-pre.0
```
f-f, line 3: I am surprised this parses, it should not 😄

anttih: Should I change the versioning scheme for prereleases? For example just stick with bumping the minor version.

f-f: Yeah, it does not - how is CI passing??

image

anttih: It does work locally for me as well with spago `0.93.14`

f-f: Ah I see - the version field needs to be inside a `publish` block. So right now the version that is returned by the `BuildInfo` should be 0.0.0

f-f:
```suggestion
  publish:
    version: 1.0.1
```

anttih: So should we then stick with bumping the patch version for prereleases or is there another versioning that spago supports for prereleases?

f-f: No support for prereleases. I think we are open to discussing adding support, but it should happen [in the registry first](https://github.com/purescript/registry-dev)

f-f: SemVer is terrible at prereleases anyways, so we'd need to diverge from it I think

anttih: Ok, I removed the prerelease script and using `1.0.1` for the next one.

f-f (approved):

PR #165, opened 2023-09-30
Add run cmd

anttih:
Adds a `run` command that allows you to skip the bundling and just eval `main`.

Adding this mainly to make running test suites in libraries easier. I was considering also adding a `test` command that would compile and run, but since spago can't be used to call `purescm run` and because I'd like to still use `spago` as much as possible the easiest path for now seems to just add a `run` and then in libraries do `spago build && purescm run --main Test.Main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,13 +24,21 @@ Then to compile your `purescm` project you can run:
 spago build
 ```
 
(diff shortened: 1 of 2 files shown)


f-f (approved):

f-f:
This is good. We should add support for `spago run`

PR #167, opened 2023-09-30
Add a package set

anttih:


anttih:
`prelude` will soon have a `spago.yaml` and then I can remove the `dependencies` field for it.

test-snapshots/spago.yaml
```diff
@@ -15,53 +15,4 @@ workspace:
...
-      ref: 8a993da0d328f72167a4742e0e2a428a790510bc
-      dependencies:
-        - functions
-        - prelude
-        - unsafe-coerce
-    unsafe-coerce:
-      git: https://github.com/purescm/purescript-unsafe-coerce.git
-      ref: 20dbd7797e985b631459b8dbd4072cf3096ee1da
-      dependencies: []
+    url: https://raw.githubusercontent.com/purescm/purescm/package-set/package-sets/1.0.0.json
```
anttih, line 18: I was thinking I could just point this to `master` before merging.

anttih: Oh but then it won't build :thinking:

f-f: Pointing to master is a questionable choice in any case - sure, we keep a hash of the package set, but it would make for a somewhat unpredictable upgrade. We should use a commit hash rather than a branch

anttih: I cannot point to a commit-ref that exists in `master`, but now just pointing to a commit in this branch.

f-f: Yeah, that's good - it should stay available even after we delete the branch

anttih: Actually this probably won't work. I just need to make another PR once this one is merged to point to a proper ref.

f-f (approved):

test-snapshots/spago.yaml
f-f: This is all a little dull actually - there's no reason why one should not be allowed to point to a filesystem path 😄

anttih: Yeah, I was looking for this in the spago docs :)

f-f: I opened https://github.com/purescript/spago/issues/1031

PR #169, opened 2023-10-02
Add packages up to `newtype`

anttih:
Linking directly to `purs` package-set.


package-sets/1.0.0.json
```diff
@@ -65,6 +65,21 @@
...
+        "unsafe-coerce"
+      ]
+    },
+    "newtype": {
+      "git": "https://github.com/purescript/purescript-newtype.git",
+      "ref": "0a0563efca18fe6b2e9d4cc0b8574e507dee4453",
+      "dependencies": [
+        "prelude",
+        "safe-coerce"
+      ]
```
f-f, line 82: We should be able to just pull upstream things from the registry:
```suggestion
    },
    "safe-coerce": "2.0.0",
    "newtype": "5.0.0"
```

f-f (approved):
Amazing work with all this FFI ❤️

PR #171, opened 2023-10-04
Add control, invariant, maybe to the package-set

anttih:


anttih:
Should we also add these as dependencies to the test suite just to check that the package-set compiles?

f-f:
> Should we also add these as dependencies to the test suite just to check that the package-set compiles?

Ah yes, great idea

anttih:
Actually I can't do that in this PR yet. Waiting for the new spago for that.

anttih:
I'd like to merge this, I need it for other package updates. I can add the test deps later.

f-f (approved):

PR #172, opened 2023-10-07
Update console, compile all packages in tests

anttih:


f-f (approved):

f-f:
This is good - I just released `0.93.15` that has the fix for local package sets, but Nix doesn't want to build it (see https://github.com/thomashoneyman/purescript-overlay/issues/47), so we can't have it in CI here yet

f-f:
@anttih until we ship this with Nix maybe you can use it locally to check that the new package set works with the test project before putting the remote URL there

anttih:
> @anttih until we ship this with Nix maybe you can use it locally to check that the new package set works with the test project before putting the remote URL there

Not sure I understand.

f-f:
Nevermind, I'll sort this nix thing out 😄

PR #173, opened 2023-10-07
Fix output flag to work with spago

anttih:
`spago` passes the output flag to a backend as `--output`.


anttih:
I also changed the flag for `bundle-app` just to be consistent.

f-f (approved):

PR #182, opened 2023-10-10
Add foldable-traversable to package set

anttih:


package-sets/1.0.0.json
```diff
@@ -41,6 +41,24 @@
...
+        "control",
+        "either",
+        "functors",
+        "identity",
+        "maybe",
+        "newtype",
+        "orders",
+        "prelude",
+        "tuples"
+      ]
```
f-f, line 60: You do not need to list the dependencies if the upstream has a `spago.yaml` file

```suggestion
      "ref": "9b265472a9e8a2cbcf21a756b0d7e584eb2c6fdf"
```

f-f (approved):

package-sets/1.0.0.json
anttih: Ah I see, I was wondering whether I needed those. I removed a bunch of those now.

f-f: Nice! 🙂

PR #184, opened 2023-10-11
Add `unfoldable` to the packageset

postsolar:


package-sets/1.0.0.json
```diff
@@ -89,6 +89,17 @@
     "safe-coerce": "2.0.0",
     "tuples": "7.0.0",
     "type-equality": "4.0.1",
+    "unfoldable": {
+      "git": "https://github.com/purescm/purescript-unfoldable.git",
+      "ref": "1cc7559e3d23537b02199d589c762087e0442f00",
+      "dependencies": [
```
anttih, line 95: You don't need the dependencies because the package has a `spago.yaml`.

postsolar: Wow, I didn't know that, cool

anttih (approved):

anttih:
Thanks!

PR #188, opened 2023-10-13
Fix bug with PrimEffect

anttih:
Fix bug where all previous bindings in a do block were dropped if the last `bind` was a `PrimEffect`.

### DIFF
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -332,23 +332,23 @@ codegenChain chainMode codegenEnv = collect []
     _ ->
       S.unthunk $ codegenExpr codegenEnv expression
 
-  finish :: Boolean -> Array _ -> NeutralExpr -> ChezExpr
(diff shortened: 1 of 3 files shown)


src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -332,31 +332,31 @@ codegenChain chainMode codegenEnv = collect []
...
 
   collect :: Array _ -> NeutralExpr -> ChezExpr
   collect bindings expression = case unwrap expression of
     Let i l v e' ->
       collect (Array.snoc bindings $ Tuple (toChezIdent i l) (codegenExpr codegenEnv v)) e'
     EffectPure e' | chainMode.effect ->
-      finish false bindings e'
+      finish false bindings (codegenExpr codegenEnv e')
     PrimEffect e' | chainMode.effect ->
-      codegenPrimEffect codegenEnv e'
```
anttih, line 351: The bug is here: we drop the `bindings` on the floor and stop.

JordanMartinez (approved):

PR #193, opened 2023-10-21
Add SRFI 152

postsolar:
This one is a lightweight strings library.
It will be necessary to implement `purescript-strings` FFI and, potentially, for other purposes.


anttih:
Would it be ok to review this with the `strings` package PRs or do you need this released before you can finish the work on `strings`?

anttih:
Ah this is a draft so maybe that was the plan?

postsolar:
Yes that was the plan. By now though after I've evaluated multiple options I can say that I'll definitely need 152 and I would like to withdraw 115 in favor of [irregex](https://github.com/ashinn/irregex). It's just marginally bigger than 115 + 115's deps (3.9k vs 3.7k LOC). There's also [pregex](https://ds26gte.github.io/pregexp/index.html) which is just 700LOC but its API is limited and not suitable. IrRegex is a single file and I wonder if it'd be possible to run dead code elimination on it to reduce the size, the same way DCE is being run for builds?

postsolar:
I think we will need to fork IrRegex anyways. It raises with `error` which isn't continuable. Right now I just did search+replace on it, seems to work fine.

anttih:
> Yes that was the plan. By now though after I've evaluated multiple options I can say that I'll definitely need 152 and I would like to withdraw 115 in favor of [irregex](https://github.com/ashinn/irregex). It's just marginally bigger than 115 + 115's deps (3.9k vs 3.7k LOC). There's also [pregex](https://ds26gte.github.io/pregexp/index.html) which is just 700LOC but its API is limited and not suitable. IrRegex is a single file and I wonder if it'd be possible to run dead code elimination on it to reduce the size, the same way DCE is being run for builds? Just for example this library supports both the POSIX/PCRE regex syntax and S-exp regex syntax. I use none of the latter.

I think what matters is that `purescm bundle-app` does this, so I don't think we need to decrease the `purescm` package size.

> I think we will need to fork IrRegex anyways. It raises with error which isn't continuable. Right now I just did search+replace on it, seems to work fine.

When does it raise a non-continuable?

postsolar:
> When does it raise a non-continuable?

Always, as in, it just uses `error` everywhere to indicate anything wrong, including incorrect user input like an attempt to construct `"+"` regex.

postsolar:
@f-f While @natefaubion had his take on `purescript-language-cst-parser` specifically, I agree with looking for a more powerful solution to avoid more issues down the road. I think the easiest option right now is some C bindings, if it's possible to plug it into the `purescm` distribution. Some other options I could think of (regardless of how realistic they are): a) forking and extending IrRegex b) porting a library from another LISP (Racket?).

postsolar:
Racket standard library's regex has [good unicode categories support](https://github.com/racket/racket/blob/f81671d6c9776ec970d5af6e39c21d8672e5652f/racket/src/regexp/parse/unicode.rkt#L31).

anttih (approved):

f-f (approved):

PR #194, opened 2023-10-21
Add SRFI 115

postsolar:
This one is for regular expressions. Will be necessary for FFI implementation of `purescript-strings`. This one has quite a number of dependencies but it seems to be the only SRFI that deals with regex.

Update:
It appears this one doesn't support traditional regex and instead only works with S-exp regex. This is neat but is a major mismatch for the PS API. Writing a translation would be too expensive, so I'm looking at alternative libraries. There doesn't seem to be any other regex SRFI library, can we add a non-SRFI one?


postsolar:
Closing. Related discussion in #193.

PR #200, opened 2023-10-31
Compile `Data.List` as scheme lists

### LINKED ISSUE #192: Output Data.List using native lists
AFAIK, this should be possible.

anttih:
Changes `Data.List` representation to a native scheme list.

This only swaps the constructors for native scheme ones, but doesn't yet include inlining rules for any other primitive list operations like folds and maps. I think we can do those separately.

Fixes #192


anttih:
Summoning @monoidmusician. Are there cases that I'm missing?

anttih:
This already passes the `lists` test suite.

anttih:
I think I'm at least missing the partially applied case for `Cons`.

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -150,6 +150,17 @@ codegenTopLevelBinding codegenEnv (Tuple (Ident i) n) =
...
+    -- (lambda (x) (lambda (xs) (scm:cons x xs)))
+    S.List
+      [ S.Identifier (scmPrefixed "lambda")
+      , S.List [ S.Identifier "x" ]
+      , S.List
+          [ S.Identifier $ scmPrefixed "lambda"
+          , S.List [ S.Identifier "xs" ]
+          , S.List [ S.Identifier $ scmPrefixed "cons", S.Identifier "x", S.Identifier "xs" ]
+          ]
+      ]
```
anttih, line 163: We might want to add this to the runtime module.

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -179,11 +190,23 @@ codegenExpr codegenEnv@{ currentModule } s = case unwrap s of
...
+  Accessor e
+    (GetCtorField (Qualified (Just (ModuleName "Data.List.Types")) (Ident "Cons")) _ _ _ "value1" _) ->
+    S.List [ S.Identifier (scmPrefixed "cdr"), (codegenExpr codegenEnv e) ]
   Accessor e (GetCtorField qi _ _ _ field _) ->
     S.recordAccessor (codegenExpr codegenEnv e) (flattenQualified currentModule qi) field
   Update e f ->
     S.recordUpdate (codegenExpr codegenEnv e) (map (codegenExpr codegenEnv) <$> f)
 
+  CtorSaturated (Qualified (Just (ModuleName "Data.List.Types")) (Ident "Nil")) _ _ _ _ ->
+    S.List [ S.Identifier $ scmPrefixed "quote", S.List [] ]
```
JordanMartinez, line 205: Thoughts on using a symbol here? If we ever need to unsafely print some scheme code, `quote ()` may not be that helpful.

src/PureScript/Backend/Chez/Convert.purs
JordanMartinez: Why might we want to add that?

src/PureScript/Backend/Chez/Convert.purs
anttih: You mean that we could instead use `(list)` or something else? I thought this was just a scheme-y way of doing this. We also use `quote` elsewhere in codegen.

anttih: Or do you mean using `'()`?

JordanMartinez: I'm not very familiar with Scheme, and it's been some time since I was learning its syntax. Is `quote ()` what Schemers typically use to indicate the end of singly-linked lists?
Initially, I was thinking we should have a value that differs from the Scheme list to help us know when this value is from Scheme code (e.g. FFI) and when it's from PureScript code. However, if we're using native lists, then I guess that's dumb. We'd want to use what Scheme users would use.

f-f: Yes, `'()` (where `'` is equivalent to `quote`) is the idiomatic way to express the end of the list.

From the repl:
```scheme
> (cons 'a (cons 'b (cons 'c '())))
(a b c)

```

JordanMartinez: Thanks for clarifying!

src/PureScript/Backend/Chez/Convert.purs
anttih: I made this change. I think it's slightly better.

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -179,11 +181,23 @@ codegenExpr codegenEnv@{ currentModule } s = case unwrap s of
     S.runUncurriedFn
       (S.Identifier $ rtPrefixed "array-ref")
       [ codegenExpr codegenEnv e, S.Integer $ wrap $ show i ]
+  Accessor e
+    (GetCtorField (Qualified (Just (ModuleName "Data.List.Types")) (Ident "Cons")) _ _ _ "value0" _) ->
```
natefaubion, line 185: I personally would match on the field index (the last slot), rather than the field name. The compiler generates a field name, but it's kind of nonsense IMO.

anttih: Perfect. I was wondering whether I can trust this name or not.

MonoidMusician:
You can also add special inlining for `List.fromFoldable [some, array, Literal]` and/or `Array.toUnfoldable [some, array, Literal]` if you want, to generate a scheme list literal directly. You'll have to look at how they codegen by default and then find a way to match on them. (It's harder since you're not gated by foreign definitions.)

anttih:
I did some benchmarking comparing `Data.List` with:

1. no compiler support (without changes in this PR)
2. with compiler support
3. with compiler support + native `foldr` and `map` implementations

Here are the results

## 1. No compiler support:

```
map: list (100000 elems)
mean   = 2.06 ms
stddev = 815.27 μs
min    = 1.13 ms
max    = 8.16 ms

foldr: list (100000 elems)
mean   = 3.40 ms
stddev = 512.92 μs
min    = 2.57 ms
max    = 8.04 ms
```

## 2. Native scheme lists:

```
map: list (100000 elems)
mean   = 994.34 μs
stddev = 391.05 μs
min    = 676.52 μs
max    = 4.66 ms

foldr: list (100000 elems)
mean   = 2.35 ms
stddev = 408.58 μs
min    = 1.95 ms
max    = 6.48 ms
```

## 3. Native scheme lists with native `foldr` and `map`

```
map: list (100000 elems)
mean   = 495.70 μs
stddev = 262.70 μs
min    = 377.78 μs
max    = 2.53 ms

foldr: list (100000 elems)
mean   = 314.36 μs
stddev = 110.30 μs
min    = 284.72 μs
max    = 2.48 ms
```

Conclusion: native lists are fast and looks like we should fork `lists` and swap out the implementations for at least `foldr` and `map`.

anttih:
PR to add native `fold`, `foldl` and `map` to `lists` https://github.com/purescm/purescript-lists/pull/1

f-f (approved):
This is seriously FAST 🚀 
Amazing work @anttih!!

JordanMartinez:
The conclusion I reach is that whatever we use to encode ADTs (records? hashmaps?) is slow. Just a thought, but if ADTs were encoded using tagged native lists (or arrays, similar to what we do for `purs-backend-es), does that produce a speed up as dramatic as this? For example:
```
data T = A | B Int | C String String
```
would produce values of
```
("A")
("B", intValue)
("C", strValue1, strValue2)
```

anttih:
> The conclusion I reach is that whatever we use to encode ADTs (records? hashmaps?) is slow. Just a thought, but if ADTs were encoded using tagged native lists (or arrays, similar to what we do for `purs-backend-es), does that produce a speed up as dramatic as this? For example:
> 
> ```
> data T = A | B Int | C String String
> ```
> 
> would produce values of
> 
> ```
> ("A")
> ("B", intValue)
> ("C", strValue1, strValue2)
> ```

Not sure if list would be good there as it still has linear time lookups. `vector` however could work, but not sure if it’s any faster than record.

PR #205, opened 2023-11-20
Fix primitive assignment effect

### LINKED ISSUE #204: Codegen/inlining issue with `Effect.Ref`
This program typechecks but fails to run:

```purescript
import Data.Array as Array
import Data.Traversable (for_)
import Effect (Effect)
import Effect.Console (log)
import Effect.Ref as Ref

main :: Effect Unit
main = do
  counter <- Ref.new 0
  for_ (Array.range 0 20) \_ -> do
    newCounter <- Ref.modify (_ + 1) counter
    log $ "New counter is " <> show newCounter
```

The error returned is:
```
Exception in number->string: #<void> is not a number
```

This is not a bug in the implementation of `Effect.Ref` - the definition for `Ref.modify` is:
```purescript
modify :: forall s. (s -> s) -> Ref s -> Effect s
modify f = modify' \s -> let s' = f s in { state: s', value: s' }
```

And if we replace the `modify` with its implementation, the program works fine:
```purescript
main :: Effect Unit
main = do
  counter <- Ref.new 0
  for_ (Array.range 0 20) \_ -> do
    let f = (_ + 1)
    newCounter <- Ref.modify' (\s -> let s' = f s in { state: s', value: s' }) counter
    log $ "New counter is " <> show newCounter
```

..which seems to point to a codegen/inlining issue.

The implementation of `modify'` is a foreign function:
```clj
  (scm:define modifyImpl
    (scm:lambda (f)
      (scm:lambda (ref)
        (scm:lambda ()
          (scm:let* ([t (f (scm:unbox ref))]
                     [v (scm:hashtable-ref t "state" (scm:quote Effect.Ref:modifyImpl-CANT-GET-STATE))])
            (scm:set-box! ref v)
            v)))))
```

Looking at the generated code, things don't seem right - the first (broken) case inlines the implementation of the foreign call, and skips the bit that reads the state out of the record (it does the unboxing but then it reboxes the wrong thing?):
```clj
  (scm:define main
    (scm:lambda ()
      (scm:let ([counter0 (scm:box 0)])
        (((for_ (Data.Array.rangeImpl 0 20)) (scm:lambda (v1)
          (scm:lambda ()
            (scm:let*
              ([_2 (scm:unbox counter0)]
               [newCounter3 (scm:set-box! counter0 (scm:fx+ _2 1))])
                ((Effect.Console.log (scm:string-append "New counter is " (Data.Show.showIntImpl newCounter3))))))))))))
```

While the correct one does not inline the foreign call:
```clj
  (scm:define main
    (scm:lambda ()
      (scm:let ([counter0 (scm:box 0)])
        (((for_ (Data.Array.rangeImpl 0 20)) (scm:lambda (v1)
          (scm:let ([_2 ((Effect.Ref.modifyImpl (scm:lambda (s2)
            (scm:let ([s$p3 (scm:fx+ s2 1)])
              (rt:make-object (scm:cons "state" s$p3) (scm:cons "value" s$p3))))) counter0)])
            (scm:lambda ()
              (scm:let ([newCounter3 (_2)])
                ((Effect.Console.log (scm:string-append "New counter is " (Data.Show.showIntImpl newCounter3)))))))))))))
```

Adding inlining suggestions to `Effect.Ref` (like `-- @inline Effect.Ref.modifyImpl never`) doesn't seem to help, so I suspect something's up with our codegen

anttih:
We need to return the value in addition to just setting the value of the box. This is not a super nice fix. I think we should refactor the prim effect stuff so that we get more control over what bindings we create, then we can also compile these to plain unboxed assigment.

Fixes #204

### DIFF
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -492,4 +492,11 @@ codegenPrimEffect codegenEnv = case _ of
     S.app (S.Identifier $ scmPrefixed "unbox") (codegenExpr codegenEnv r)
   EffectRefWrite r v ->
     S.List
-      [ S.Identifier $ scmPrefixed "set-box!", codegenExpr codegenEnv r, codegenExpr codegenEnv v ]
+      [ S.Identifier $ scmPrefixed "begin"
+      , S.List
+          [ S.Identifier $ scmPrefixed "set-box!"
+          , codegenExpr codegenEnv r
+          , codegenExpr codegenEnv v
+          ]
+      , S.List [ S.Identifier $ scmPrefixed "unbox", codegenExpr codegenEnv r ]
+      ]
--- a/test-snapshots/src/snapshots-input/Snapshot.EffectRef.purs
+++ b/test-snapshots/src/snapshots-input/Snapshot.EffectRef.purs
@@ -4,7 +4,8 @@ module Snapshot.EffectRef where
 import Prelude
 
 import Effect (Effect)
-import Effect.Ref (Ref, modify_, new, read)
+import Effect.Ref (Ref, modify, modify_, new, read)
+import Effect.Class.Console (log)
 import Test.Assert (assert)
 
 positionZero :: Effect (Ref Int)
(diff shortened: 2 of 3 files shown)


anttih:
I guess we don't need to unbox it right after `box-set!` and could just return the value that was set.

PR #206, opened 2023-12-07
Custom immutable, persistent string implementation

anttih:
This is my experiment which turned into an actual implementation of a custom string type written in scheme. I call the new type `pstring` as in "Persistent" or "Ps" string.

The goal of this is to try to match the performance characteristics of JS strings, which means constant-time slicing.
The string type implemented here is a UTF-16 encoded, immobile bytevector with a length and offset.
It uses structural sharing to avoid copying when slicing (the length and offset parts).

There have been multiple SRFIs ([srfi-130](https://srfi.schemers.org/srfi-130/srfi-130.html), [srfi-135](https://srfi.schemers.org/srfi-135/srfi-135.html)) that try to solve this same problem, but they are not widely implemented because of the difficulty to support them everywhere, a problem that we (hopefully) don't have. It's mostly slightly annoying to have to wrap native strings with `string->pstring` when writing FFI or when using the REPL.

The implementation is quite naive, in that a very small string slice can prevent a huge buffer from being garbage collected. Hopefully this is something we can live with for now as a starting point.

~~Each literal also needs to be wrapped in `string->bytestring` and I don't know the performance implications of that (yet).~~ **Update**: not anymore as there is a macro that evaluates the literals to bytevectors at compile time.

~~Record keys are now also bytestrings.~~
**Update**: record keys are symbols, but this was already done in a separate PR.

I kept the encoding as UTF-16 at least for now to match the JS implementation. If and when PS strings get less coupled with UTF-16 encoding (@MonoidMusician's [proposal](https://cofree.coffee/~verity/purs_unicode.html)), we can change the encoding to UTF-8.

**Note** we don't allow lone surrogates in strings. This is because we still need to go through Chez strings in some places and those cannot contain lone surrogates.

~~[PR for `strings`](https://github.com/purescm/purescript-strings/pull/2)~~
[PR for core libs](https://github.com/purescm/purescript-core/pull/4)


anttih:
One potential show stopper here is regex. Firstly, `irregex` doesn't provide full PCRE compatiblity, for example it doesn't support the `\p` patterns. Secondly parsing with PS CST parser is very slow, I'll need to take a closer look at what is causing the slowness.

anttih:
> I had a look at our irregex file and I noticed that everything is using top-level defines there - [the Chez documentation explicitly instructs against using top-level define](https://cisco.github.io/ChezScheme/csug10.0/use.html#./use:h6), as it prevents the runtime from inlining: I suspect that's the cause of the slowness here.

The file that we include is [this](https://github.com/purescm/purescm/blob/1a4ada1b94dc069baa9291aed6adba29f58d1061/vendor/purs/runtime/irregex.ss), so the code is wrapped in a library, which to my understanding does conform to this:

> Fortunately, it is easy to restructure a program to avoid top-level bindings. This is naturally accomplished for portable code by placing the code into a single RNRS top-level program or by placing a portion of the code in a top-level program and the remainder in one or more separate libraries. Although not portable, one can also put all of the code into a single top-level module form or let expression, perhaps using include to bring in portions of the code from separate files. The compiler performs some optimization even across library boundaries, so the penalty for breaking a program up in this manner is generally acceptable. The compiler also supports whole-program optimization (via compile-whole-program), which can be used to eliminate all overhead for placing portions of a program into separate libraries.

anttih:
>  Secondly parsing with PS CST parser is very slow, I'll need to take a closer look at what is causing the slowness.

One major cause of slowness is our type class dictionaries which use string hashtables. I have a WIP where I use symbols as keys which improves the performance significantly. I'll make a PR of that.

Then I will investigate using PCRE for regex in order to support all of PCRE and to see if that also improves performance even more.

f-f (approved):
Left a small request for docs, but otherwise this looks great - excellent stuff @anttih! 👏

f-f (changes requested):
I tested this out on my macos, and I can't get the nix-shell to run as committed - I removed the patching of the chez executable, and added this to the `mkShell` instead:
```
shellHook = ''
              export DYLD_LIBRARY_PATH="$DYLD_LIBRARY_PATH:${pkgs.pcre2}/lib";
            '';
```

`DYLD_LIBRARY_PATH` only works on macOS, Linux will need `LD_LIBRARY_PATH` - we should add a macOS to the CI to make sure that changes are somewhat portable

f-f:
@toastal since you built purescm for nixpkgs: how would this `pcre2` dependency work there?

f-f:
That did it! 🎉 merging.

f-f (approved):

PR #207, opened 2023-12-08
Implement runtime module in Scheme

### LINKED ISSUE #147: Implement runtime module in scheme?
Is there any particular reason why the [runtime module](https://github.com/purescm/purescm/blob/master/src/PureScript/Backend/Chez/Runtime.purs) is implemented in PureScript and not in Scheme? If not, I'd like to propose we could implement it in scheme for the following reasons:

- ~~We need an extra build step to copy the generated scheme file to the `purs-backend-chez` runtime directory. If the file is just a scheme file we can easily just place it alongside the srfi runtime libs. This helped me get a POC of the build tool going.~~
- It does provide some type safety to be written in PureScript, but not much.

anttih:
Takes the compiled scheme file and places it in the `vendor/runtime.ss`. Not this is a breaking change if a library imports the file with the `lib` suffix. I see no reason to keep the `lib` so I removed it.

Fixes #147


test-snapshots/src/snapshots-output/Snapshot.List.ss
```diff
@@ -25,10 +25,10 @@
     (scm:cons 1 (scm:quote ())))
 
   (scm:define curried
-    rt:cons)
+    rt:kons)
 
   (scm:define cons2
-    (rt:cons 2))
+    (rt:kons 2))
```
JordanMartinez, line 31: I'm guessing the reason why this is being renamed is to make it that much clearer that this is not the scheme `cons`?

natefaubion: `list-cons`?

anttih: > I'm guessing the reason why this is being renamed is to make it that much clearer that this is not the scheme cons?

Yes, and I removed the prefix `scm:` from the `chezscheme` import which introduced this name clash. I didn't want to bother doing renamed imports for this single name

`list-cons` is obviously a better name. I'll use that. Thanks!

anttih:
Build is failing because this requires changes in the libs.

f-f (approved):
I think this is a good change, the new runtime module seems easier to reason about.

anttih:
I will need to merge this first, then make a release, then make PRs for the libs.

anttih:
Actually merging is blocked by CI. What I can do is point the package refs in the tests to unmerged PRs. Should I do that? Then I'll just have to update those refs again when the PRs are merged.

JordanMartinez:
I think you should do whatever is easiest for you since you're the one doing most of this work.

PR #214, opened 2023-12-19
Use symbols as record labels

anttih:
While profiling the `strings` library I found that string hashing is our biggest hot-spot. We repeatedly hash the same strings over and over in many places. This changes records to use symbols as labels instead of strings. Using symbols means there is no hashing and we can use fast `eq?` checks for key lookups. This improved my particular parsing benchmark result around 60%.

A few implementation details:
- We no longer need SRFIs 125, 126 and 128, and so they have been removed. We just use `symbol-hashtable-*` functions provided by Chez.
- We use the value `(cons #f #f)` as the value that will be returned internally in the case the key is not found, because `symbol-hashtable-ref` needs a default value. Note that you can still store that value in a record because we use `eq?` reference check to check for that particular value.
- All record labels are wrapped in `string->symbol` instead of doing escaping on our own. I started implementing the escaping but then realized that Chez inlines this at compile time:

```
> (expand/optimize '(string->symbol "foo bar"))
'foo\x20;bar
```

When we are self-hosting we can do the escaping at purescm compile time and get nicer compiler output, but this makes the most sense right now.


anttih:
If we implement a custom string type then `string->symbol` won't inline, and we need to do at least some escaping to avoid most of the extra allocations. We can deal with that when we get there.

f-f (approved):

f-f:
> If we implement a custom string type then `string->symbol` won't inline

Even if we add a `bytestring->symbol`? Is that because `string->symbol` is implemented in C?

anttih:
> > If we implement a custom string type then `string->symbol` won't inline
> 
> Even if we add a `bytestring->symbol`? Is that because `string->symbol` is implemented in C?

TBH I'm not entirely sure why it doesn't inline, I'll have to dig deeper. What currently happens in my `bytestring->symbol` is `(string->symbol (bytestring->string "foo"))` and the part that probably doesn't inline is the decoding from bytevector to string.

PR #229, opened 2024-02-16
Return with correct exit code when compilation fails

anttih:
This was a recent change which I'm now reverting. Basically this fixes the case when there is an error during compilation, either when AOT compiling or when compiling as part of the `run` command. I had forgotten why I put these error handlers in there, but it was this case. We should probably add a test suite for the CLI at some point.


f-f (commented):
Let's add a comment that explains why these should stay here so we don't move them again 😄

f-f (approved):

PR #234, opened 2024-02-19
Dump version to 1.8.2 and add release script

anttih:
Moved some logic from `package.json` to a shell script because it's more easier to understand the steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Install requisites:
 You can install purescm with npm:
 
 ```
-npm i --global purescm@latest
+npm i --global purescm
 ```
 
 ## Usage
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Install requisites:
 You can install purescm with npm:
 
 ```
-npm i --global purescm@next
+npm i --global purescm@latest
 ```
 
 ## Usage
--- a/publish-release.sh
+++ b/publish-release.sh
@@ -8,9 +8,10 @@
 
 set -eu
 
-if [ -z "$1" ] then
-  echo "No arguments supplied"
-  exit 1
+if [ $# -eq 0 ]
+  then
+    echo "No arguments supplied"
+    exit 1
 fi
 
 # bump npm version and tag commit
--- a/publish-release.sh
+++ b/publish-release.sh
@@ -26,9 +26,10 @@ git pull
 # bump npm version
 npm version $1 --no-git-tag-version
 
+# `npm version` also updates spago.yaml, so we grap the new
+# tag name from there
 NEW_TAG=$(grep -Po "(?<=version: )([\d.]+)" spago.yaml)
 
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -10,7 +10,7 @@ Install requisites:
 You can install purescm with npm:
 
 ```
-npm i --global purescm@next
+npm i --global purescm@latest
```
f-f, line 13: Latest is default:
```suggestion
npm i --global purescm
```

publish-release.sh
```diff
@@ -0,0 +1,27 @@
...
+#
+
+set -eu
+
+if [ $# -eq 0 ]
+  then
+    echo "No arguments supplied"
+    exit 1
+fi
+
```
f-f, line 16: We should also check that we're not publishing from a random branch:
```suggestion

# Fail if we are not on trunk
BRANCH=$(git branch --show-current)
if [ "${BRANCH}" != "master" ]; then
  echo "Please checkout master branch"
  exit 1;
fi
git pull

```

publish-release.sh
```diff
@@ -0,0 +1,27 @@
...
+# bump npm version and tag commit
+npm version $1
+
+# bump also in spago
+git add spago.yaml
+git commit --amend --no-edit
```
f-f, line 22: Why the `--amend --no-edit`?

publish-release.sh
```diff
@@ -0,0 +1,27 @@
...
+
+rm -rf bundle output output-es
+npm run build
+
+npm publish --tag latest
```
f-f, line 27: Should we also push the new commit?

```suggestion
npm publish --tag latest

git push
```

publish-release.sh
anttih: My workflow is/was that I check out the release branch, then publish (or make a PR first, then publish). So I'm actually never in master. I thought that we can't commit straight to `master`, not sure if we have set it up as protected though.

publish-release.sh
anttih: I wanted to amend the commit that `npm version` makes. I could just make another commit, or I can prevent `npm version` not do the commit, but then I need to also do a `git tag` here (which `npm version` does automatically).

publish-release.sh
anttih: Sure if pushing to master is what we want.

publish-release.sh
f-f: We can commit straight to master (because we have the admin bit) and I'd prefer it for release commits, so we don't have to go through the bureaucracy of a branch/merge just for bumping a release number.

Trunk should always be release-ready anyways, so we're all set for this kind of workflow.

anttih: Makes sense.

f-f (approved):
👏

PR #235, opened 2024-05-10
Optimize string operations by using a "cursor"

anttih:
This optimizes many of the string operations by using a new abstraction called _cursor_ that is much like a Scheme port but that works better for our use cases. Cursor has an imperative API that reduces allocations significantly compared to unconsing.

While removing the uses of `pstring-uncons-code-unit` I also found a bug in both `pstring-index-of` and `pstring-last-index-of` where the match wasn't found at all if the string and pattern shared a longer prefix.

This also changes how we decode code points from the UTF-16 bytevector: we now replace lone surrogates with the Unicode Replacement Character. This way we can implement `cursor-read-char` in terms of `cursor-read-code-point`.

Why not use Scheme ports for fast iteration over String? We could use [`make-custom-textual-input-port`](https://scheme.com/tspl4/io.html#./io:s42) but it's API assumes you want to allocate a Scheme string every time you read from the port. This would work since we only ever need the first character from the port, but is not optimal for this use case. We also don't currently need the `port` abstraction for anything.

Fix for `strings` https://github.com/purescm/purescript-core/pull/10

### DIFF
--- a/vendor/purs/runtime/pstring-buffer.ss
+++ b/vendor/purs/runtime/pstring-buffer.ss
@@ -101,6 +101,5 @@
               ;; one-word encoding
               [else (begin (string-set! out char-i (integer->char w1)) (loop (fx+ i 1) (fx1+ char-i)))]))
           (begin (string-truncate! out char-i) out)))))
-
   )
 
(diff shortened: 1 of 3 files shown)


anttih:
Just a quick benchmark for `String` equality comparing two large strings 100 times (without no optimizations turned on):

Old:

```
    196 collections
    0.598815825s elapsed cpu time, including 0.008619467s collecting
    0.601080700s elapsed real time, including 0.009075998s collecting
    1649015104 bytes allocated, including 1652884528 bytes reclaimed
```

New:

```
    no collections
    0.236853978s elapsed cpu time
    0.237545801s elapsed real time
    6400 bytes allocated
```

f-f (approved):
Amazing work ❤️

PR #238, opened 2024-06-12
Print a stack trace

anttih:
Prints a stack trace for caught exceptions. It will print something like this:

```
Exception: attempt to apply non-procedure #f
  at flexvector-map/index-into! (/home/antti/code/purescm/purescm/vendor/purs/runtime/srfi/:214/impl.sls:322,25)
  at f (/home/antti/code/purescm/purescm/vendor/purs/runtime/srfi/:214/impl.sls:340,5)
  at pstring-regex-replace-single (/home/antti/code/purescm/purescm/vendor/purs/runtime/pstring.ss:746,29)
  at testStringRegex (output/Test.Data.String.Regex/lib.ss:74,50)
  at main (output/Test.Data.String.Main/lib.ss:33,15)
```

### DIFF
--- a/src/Main.purs
+++ b/src/Main.purs
@@ -202,8 +202,10 @@ runBundle cliRoot args = do
     mainContent = String.joinWith "\n"
       [ "#!chezscheme"
       , "(import (chezscheme)"
+      , "        (only (purs runtime stack-trace) print-stack-trace)"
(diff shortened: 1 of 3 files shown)


f-f (approved):
Ah, wonderful

PR #244, opened 2024-07-02
Fix string-truncate with 0 length

f-f:
`string-truncate` will panic if given a length of 0, so we skip it if that's the case

### DIFF
--- a/vendor/purs/runtime/pstring-buffer.ss
+++ b/vendor/purs/runtime/pstring-buffer.ss
@@ -100,6 +100,8 @@
                  (loop (fx+ i 1) (fx1+ char-i)))]
               ;; one-word encoding
               [else (begin (string-set! out char-i (integer->char w1)) (loop (fx+ i 1) (fx1+ char-i)))]))
-          (begin (string-truncate! out char-i) out)))))
+          (if (eq? len 0)
+            out
+            (string-truncate! out char-i))))))
   )
 


anttih:
I think this might be hiding a bug elsewhere. It should not happen that `char-i` is greater than `len` when we truncate, which I think is the issue here.

f-f:
It will likely become apparent once I find the original repro

f-f:
I have a repro with the `Foreign.Object` quickcheck tests:

```
len: 0
char-i: 0
string-length: 0
out: 
Exception in string-truncate!: "" is not a mutable string
  at _lookup (/Users/fabrizio/code/purescm/vendor/purs/runtime/pstring.ss:523,21)
  at --main-- (output/Test.Foreign.Object/lib.ss:437,26)
  at --main-- (output/Test.QuickCheck/lib.ss:92,74)
  at --main-- (output/Control.Monad.State.Trans/lib.ss:111,16)
  at _4 (output/Test.QuickCheck/lib.ss:115,42)
  at go5 (output/Test.QuickCheck/lib.ss:124,60)
  at --main-- (output/Test.QuickCheck/lib.ss:128,29)
  at objectTests (output/Test.Foreign.Object/lib.ss:431,16)
```

f-f:
The issue doesn't show up with running an optimised build, with `(optimize-level 3)`

anttih:
Can be reproduced in the Chez repl:

```
> (string->symbol "")
> (string-truncate! (make-string 0) 0)
```

f-f:
But it's fixed on latest:
```
Chez Scheme Version 10.1.0-pre-release.1
Copyright 1984-2024 Cisco Systems, Inc.

> (string->symbol "")
||
> (string-truncate! (make-string 0) 0)
""
> 
```

f-f:
Fixed by #246

PR #246, opened 2024-07-03
Upgrade Chez and other packages in nix config

anttih:


f-f:
We need to stay on the racket fork, or upgrade to 10.0 to support things like ARM macs

anttih:
Seems like the issue linked in `flake.nix` is now closed so using nixpkgs-unstable which gives us Chez 10.x.

f-f (approved):
Wonderful!

PR #250, opened 2024-08-19
Use libicu for pstring-upcase and pstring-downcase

### LINKED ISSUE #249: Use ICU for string case conversions
Currently `String.toUpper` is two orders of magnitude slower in purescm than in node. We are doing extra allocations because we need to first convert to a Chez string, then do the conversion and then convert back to pstring. Proper unicode aware case conversion is not trivial to do so I was thinking that maybe we just add bindings to libicu.

anttih:
Use libicu for case conversions. No error handling yet, and uses the "default" locale.

```
Node:
String.toUpper: { mean = 3.52 μs | stddev = 5.87 μs | min = 2.28 μs | max = 61.39 μs }

Old purescm:
String.toUpper: { mean = 135.52 μs | stddev = 20.98 μs | min = 120.78 μs | max = 278.43 μs }

New purescm:
String.toUpper: { mean = 40.49 μs | stddev = 4.44 μs | min = 38.98 μs | max = 79.51 μs }
```

We are now only one order of magnitude slower than node :)

Fixes #249

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ Install requisites:
 
 * [`chezscheme`](https://cisco.github.io/ChezScheme/) - you should have the `scheme` command available in your `$PATH`.
 * [`libpcre2`](https://pcre2project.github.io/pcre2/) - `purescm` has a runtime dependency on the 16-bit variant of pcre2. Check your package manager for `pcre2` or `pcre2-16` or similar.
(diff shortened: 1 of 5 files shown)


f-f (approved):
This is great 👏

anttih:
I added error handling, and also now pass a null value to the `locale` argument which picks up the locale from your local system.

PR #251, opened 2024-08-25
Add nix caching for CI

anttih:
Cache the nix artifacts between builds (mostly Chez) and don't install node tools (we use nix after all).


anttih:
Let's see if this would avoid building Chez on every run.

anttih:
I guess it works, from 6m to 1m :)

anttih:
Also fixed errors about missing nixpkgs by using `nix develop` which picks up the nixpkgs from our flake.

PR #252, opened 2024-09-22
Allow running scheme with custom dynamic libraries on macOS

f-f:
Picking up custom paths for loading dynamic libraries through the `DYLD_LIBRARY_PATH` is [not allowed in the latest macOS version](https://developer.apple.com/library/archive/documentation/Security/Conceptual/System_Integrity_Protection_Guide/RuntimeProtections/RuntimeProtections.html) when launching processes in the shell. So we go around this by looking up a custom variable and setting `DYLD_LIBRARY_PATH` when launching `scheme` instead - this works and lets us load dynamic libraries through the envvar instead of having to patch the binary's `rpath`.


src/Main.purs
```diff
@@ -260,7 +261,18 @@ runRun cliRoot args = do
...
-  spawned <- execa schemeBin arguments identity
+
+  -- Apple blocks DYLD environment variables, so loading libraries with DYLD_LIBRARY_PATH
+  -- will not work across process spawns. We go around this by carrying the library path
+  -- in CHEZ_DYLD_LIBRARY_PATH and setting DYLD_LIBRARY_PATH when spawning. For more info see:
+  -- https://developer.apple.com/library/archive/documentation/Security/Conceptual/System_Integrity_Protection_Guide/RuntimeProtections/RuntimeProtections.html
+  env <- liftEffect $ Process.getEnv
+  case Object.lookup "CHEZ_DYLD_LIBRARY_PATH" env of
+    Just dyldPath -> do
+      liftEffect $ Process.setEnv "DYLD_LIBRARY_PATH" dyldPath
```
anttih, line 272: Why not just read and set `DYLD_LIBRARY_PATH` directly instead of reading from `CHEZ_DYLD_LIBRARY_PATH`. The only case I can think of is if you want to link to different versions of some library in the parent process and `scheme`.

anttih (commented):
Do you want to also update the flake?

src/Main.purs
f-f: We can't just read `DYLD_LIBRARY_PATH` because it will be stripped away when launching `purescm` at the command line - from the link above:
> environment variables, such as DYLD_LIBRARY_PATH, are purged when launching protected processes

f-f:
I will attempt to update the flake - last time I tried patching the dynamic library loader on linux I didn't have much success, but let's see

anttih (approved):

PR #254, opened 2024-10-07
Update deps

anttih:
Still locally it wants to create the spago.lock for the test-snapshots project everytime. It also takes quite a long time on "Reading Spago workspace configuration..." when running the test suite.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -34,11 +34,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1723891200,
-        "narHash": "sha256-uljX21+D/DZgb9uEFFG2dkkQbPZN+ig4Z6+UCLWFVAk=",
+        "lastModified": 1728217273,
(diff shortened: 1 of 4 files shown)


anttih:
@f-f Any idea what is going on?

f-f:
I'll have a look

f-f:
The spago lock being considered out of date is expected: since we are using a local package set we had the issue of Spago not picking up changes to it, and fixed it here: https://github.com/purescript/spago/pull/1201

f-f:
I also get a long wait (about 18s) on the initial phase, but in my case it's stuck at `Gathering all the spago configs in the tree...`. I'm still trying to figure out where that gets stuck

f-f:
Ah I got it - we don't have a `.gitignore` file in the `test-snapshots` folder, so Spago will happily climb through that. Adding a `.gitignore` with just that entry works, but it sounds like we could ignore some well-known locations by default (`.spago`, `node_modules`, etc)

f-f:
Patch for spago here: https://github.com/purescript/spago/pull/1293

f-f (approved):

PR #255, opened 2024-10-22
Implement lazy initialization for recursive top level bindings

anttih:
Implements lazy initialization for top-level recursive bindings. This allows mutually recursive top-level definitions to refer to each other during initialization. It also provides a better runtime error in the case that a definition is referring to itself during initialization. These rarely happen in the wild, especially since we use the optimizer that sometimes inlines away these recursive references. However, if you prevent inlining these happen more often. You can read more about lazy initialization [here](https://github.com/purescript/purescript/blob/0527e71725257bf16a3c7bd5fce608e856a77dae/CHANGELOG.d/breaking_fix-4179.md).

This does _not_ implement lazy initialization for non-top-level bindings. I'd like to do this in a separate PR. And note that the Scheme syntax `letrec*` does not allow mutually recursive references during initialization. Same goes for `define`.

### Implementation

There is a new field `lazyTopLevelRefs` that is a `Set` of names that should be initialized lazily. These value is populated for each recursive binding group and then during codegen we lookup all the qualified top-level-names from there, and if found we generate a lazy reference.

A lazy binding is a thunk, that when called will initialize the binding and keep track of self-references to provide an error message. Lazy definitions are implemented as a new special form in the runtime called `define-lazy` just to prevent noise in the output caused by the thunks.

We could use the normal `Define` constructor for the lazy definitions but then we would need to add a boolean flag to `Define` to indicate whether the binding should be exported or not (lazy bindings don't need to be exported), or we could export even the lazy bindings. I thought that a new `DefineLazy` would be the cleanest and would allow using the `define-lazy` special form more easily. That being said, I'm not convinced that generating macro forms is always a good idea, here it feels ok.

The test suite now also supports tests that fail at runtime by a test naming convention. It doesn't check the failure reason. At some point we refactored the test suite to use the production builder which makes it difficult to hook into the tests such that we could for example annotate the test source file with the failure reason. We should look into this later.

### DIFF
--- a/src/PureScript/Backend/Chez/Convert.purs
+++ b/src/PureScript/Backend/Chez/Convert.purs
@@ -8,18 +8,21 @@ import Data.Array.NonEmpty as NEA
 import Data.Array.NonEmpty as NonEmptyArray
 import Data.Bifunctor (bimap)
 import Data.Char (toCharCode)
+import Data.Filterable (separate)
+import Data.Foldable (fold, foldMap)
 import Data.Int as Int
 import Data.Maybe (Maybe(..))
 import Data.Monoid (power)
 import Data.Newtype (unwrap, wrap)
 import Data.Number (isNaN)
+import Data.Set (Set)
 import Data.Set as Set
 import Data.String as String
 import Data.String.CodeUnits as CodeUnits
 import Data.String.Regex as R
 import Data.String.Regex.Flags as R.Flags
 import Data.String.Regex.Unsafe as R.Unsafe
-import Data.Tuple (Tuple(..), uncurry)
+import Data.Tuple (Tuple(..), fst, uncurry)
 import Data.Tuple as Tuple
 import Partial.Unsafe (unsafeCrashWith)
 import PureScript.Backend.Chez.Constants (libChezSchemePrefix, moduleForeign, moduleLib, rtPrefixed, runtimePrefix, scmPrefixed, undefinedSymbol)
@@ -33,13 +36,14 @@ import Safe.Coerce (coerce)
 
 type CodegenEnv =
   { currentModule :: ModuleName
+  , lazyTopLevelRefs :: Set Ident
   }
 
 codegenModule :: BackendModule -> ChezLibrary
 codegenModule { name, bindings, imports, foreign: foreign_ } =
   let
     codegenEnv :: CodegenEnv
-    codegenEnv = { currentModule: name }
+    codegenEnv = { currentModule: name, lazyTopLevelRefs: Set.empty }
 
     definitions :: Array ChezDefinition
     definitions = Array.concat $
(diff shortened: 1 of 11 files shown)


anttih:
I've started working on incremental compilation support and there I hit this issue immediately because the optimizer doesn't inline anything and there are more recursive bindings. The more pressing issue are the top-level bindings and so the `LetRec` forms are left for later. I'm also still not sure how big of an issue this is for non-top-level bindings.

src/PureScript/Backend/Chez/Convert.purs
```diff
@@ -118,8 +122,47 @@ codegenTopLevelBindingGroup
...
+          { lazyTopLevelRefs = foldMap (Set.singleton <<< fst) lazyBindings.no }
+      fold
+        [ Array.concatMap (codegenTopLevelBinding codegenEnv') lazyBindings.yes
+        , map (codegenLazyTopLevelBinding codegenEnv') lazyBindings.no
+        , map (\(Tuple (Ident ident) _) -> Define ident $ S.forceLazyRef ident) lazyBindings.no
+        ]
+  | otherwise = Array.concatMap (codegenTopLevelBinding codegenEnv) bindings
+
+isLazyBinding :: Tuple Ident NeutralExpr -> Boolean
+isLazyBinding (Tuple _ expr) = case unwrap expr of
```
anttih, line 142: I wonder if I should be more aggressive here @natefaubion?

natefaubion: By aggressive, you mean try harder to detect more lazy bindings? It's not really an issue of correctness (you can make every binding in every recursive group lazy). You are are just potentially eliminating overhead where it's not needed. The one for the ES backend does work a little harder, yes, but not much.
https://github.com/aristanetworks/purescript-backend-optimizer/blob/bdef8b58cd8591bef7fd499c892989604e6b918d/backend-es/src/PureScript/Backend/Optimizer/Codegen/EcmaScript/Convert.purs#L802-L854

src/PureScript/Backend/Chez/Printer.purs
```diff
@@ -235,6 +238,18 @@ printDefinition = case _ of
   Define ident expr ->
     printNamedIndentedList (D.words [ D.text $ scmPrefixed "define", D.text ident ])
       $ printChezExpr expr
+  DefineLazy ident (ModuleName m) expr -> do
+    let origIdentName = Json.stringify $ Json.fromString ident
+    let moduleName = Json.stringify $ Json.fromString m
```
f-f, line 243: Why printing with json?

f-f:
I think using macros for code we generate is overall a good idea, it keeps it small and it's overally easy on the cp0 optimiser.

src/PureScript/Backend/Chez/Printer.purs
f-f: Makes sense 👍

f-f (approved):

PR #256, opened 2024-10-26
Use the same symbol for `PrimUndefined`

anttih:
We generated a new symbol each time using `gensym`, but I don't think it's necessary. Now just using `'purs-undefined`.

### DIFF
--- a/src/PureScript/Backend/Chez/Constants.purs
+++ b/src/PureScript/Backend/Chez/Constants.purs
@@ -22,6 +22,3 @@ moduleLib = "lib"
 
 schemeExt :: String
 schemeExt = ".ss"
-
-undefinedSymbol :: String
-undefinedSymbol = "purs-undefined"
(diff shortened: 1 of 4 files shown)


purefunctor (approved):
Was originally added to track unimplemented stuff I think but LGTM

anttih:
> Was originally added to track unimplemented stuff I think but LGTM

Oh I see, so it wasn't for `PrimUndefined` but for "not implemented". Then I'd like to be more explicit about the use case here and use something else than `'purs-undefined`. I updated the PR to just use `'undefined'`. I think we could also use something like `'__unused`, but not sure.

f-f:
+1 for being explicit about it being for unimplemented stuff. No preferences about the name as long as it's more explicit than it is at the moment

anttih:
> +1 for being explicit about it being for unimplemented stuff. No preferences about the name as long as it's more explicit than it is at the moment

It's not for unimplemented anymore, I think it's only used for an unused value being passed around to thunks.

anttih:
Using `'undefined` for now to map to `PrimUndefined`.

PR #257, opened 2024-10-26
Use a special runtime function for failed pattern match

anttih:
There are no other failure cases than the failed pattern match case, so instead of handling arbitrary error strings we just use a special runtime function for failed pattern matches. This also cleans up the output nicely.

This is part of a series of refactorings I'm doing to use `StringExpr` only for strings that actually need to be escaped. Then we can have the string escaping only in one place.

### DIFF
--- a/vendor/purs/runtime.ss
+++ b/vendor/purs/runtime.ss
@@ -21,7 +21,8 @@
     pstring>?
     pstring>=?
     pstring<?
-    pstring<=?)
+    pstring<=?
+    fail)
   (import
     (chezscheme)
     (only (purs runtime pstring) string->pstring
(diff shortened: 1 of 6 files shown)


vendor/purs/runtime.ss
```diff
@@ -108,4 +109,15 @@
     (lambda (x) (lambda (xs) (cons x xs))))
 
 
+  ;
+  ; Runtime failures
+  ;
+
+  (define (fail)
+    (raise
```
f-f, line 117: I sometimes see crashes about `&non-continuable`, and while I am not sure where they are from, I suspect it's because we use `raise` instead of `raise-continuable` - from the [docs](https://scheme.com/tspl4/exceptions.html#./exceptions:s3):

>  If an exception handler attempts to return to the continuation of a non-continuable exception, however, a new exception with condition type `&non-continuable` is raised

Should this one be a `raise-continuable`?

f-f (approved):

PR #260, opened 2024-11-03
Reorganize scheme libs

anttih:
The naming of things under `vendor/` has been bothering me ever since we added more of our own scheme code (pstring and runtime libraries), because now the `vendor/` contains all the scheme code whether it's vendored or not. I also think it's a bit misleading to prefix the srfi modules with "purs runtime" in the module path. So I'm trying to tackle these problems with this change.

This is what it does:

* Renames `vendor` to `lib`
* SRFI modules are renamed from `(purs runtime srfi *)` to just `(srfi *)`.
* purescm runtime modules are renamed from `(purs runtime *)` to `(purescm *)`.

After this there is a one-to-one mapping of the [chez-srfi](https://github.com/arcfide/chez-srfi) lib code to `lib/srfi`. Also when the scheme code is compiled as part of `purescm bundle-app` it gets compiled to `output/` instead of `output/vendor` which simplifies things a bit.


package-sets/1.0.0.json
```diff
@@ -5,66 +5,66 @@
   "packages": {
     "arrays": {
       "git": "https://github.com/purescm/purescript-core.git",
-      "ref": "98f12cc76d4941f979f2530da94f5b46ce9551cd",
+      "ref": "1982feebeb96103c9e8caaf6a646498445e61cb5",
```
f-f, line 8: If we used a yaml reference we could have this hash in a single place
```suggestion
      "ref": &core-ref "1982feebeb96103c9e8caaf6a646498445e61cb5",
```

See e.g. how's it done [here](https://github.com/purescript/spago/issues/1295#issuecomment-2450846910)

f-f (approved):
Thanks for making this change!

I left a comment but we don't have to implement it immediately, next time we touch that file also works

package-sets/1.0.0.json
anttih: But this is not yaml. :)

PR #265, opened 2026-02-19
Adding Bytevectors

harryprayiv:
Added bytevectors to try and get more performance out of my sha2 and sha3 purescm implementations.  It didn't help in my test cases/benchmarks, but this change passes all tests and (IMO) adds much needed functionality.  I also added a version bump to reflect this difference.

PureScript’s `Array Int` compiles to SRFI 214 flexvectors, so every byte is a boxed Scheme object. On 64-bit Chez that’s roughly 8 bytes per element. A million bytes stored as `Array Int` eats ~8 MB of heap when the same data in a bytevector would be 1 MB.

I’ve been writing crypto libraries for purescm (SHA-3, SHA-2 and HMAC) and every one of them ends up declaring its own opaque `ByteArray` backed by Chez bytevectors:

```purescript
foreign import data ByteArray :: Type
foreign import arrayToByteArray :: Array Int -> ByteArray
foreign import byteArrayToArray :: ByteArray -> Array Int
```

```scheme
(define arrayToByteArray (lambda (fv) (bv-from-flexvector fv)))
(define byteArrayToArray (lambda (bv) (bv-to-flexvector bv)))
```

This same boilerplate shows up in every library that touches bytes. And the `ByteArray` types are all incompatible at the PureScript level even though they’re literally the same Chez bytevector at runtime.

The JS backend solved this years ago with `node-buffer`. There’s one shared `Buffer` type and everyone uses it. We don’t have an equivalent.

There’s also a conversion cost that’s easy to miss. The string path is fine because `string->utf8` is a Chez builtin that returns a bytevector directly. But `Array Int` input has to go through flexvector -> bytevector (element by element copy) -> do the actual work -> bytevector -> flexvector (element by element copy back). Two full traversals just to cross the type boundary. In my crypto benchmarks this doesn’t dominate because the hash computation is the bottleneck, but for something like shuffling binary data between a database driver and a serialization library it’s pure waste.

The other issue is ecosystem interop. Every Chez library that handles binary data speaks bytevector: file I/O, sockets, CBOR, compression, database wire protocols. Without a shared type, each purescm binding has to write its own conversion layer.

`(purescm bytevector)` already has the Scheme-side utilities (`bv-from-flexvector`, `bv-to-flexvector`, `bv-to-hex`, `bv-from-hex`). What’s missing is a PureScript-level package that libraries can depend on instead of each declaring their own private `ByteArray`. That would let you pipe output from one library into another without round-tripping through `Array Int`.


f-f:
Interesting, thanks @harryprayiv!

Question: do we need this in core? My cursory reading of the patch is that this is basically a wrapper around the [existing bytevector functionality](https://scheme.com/tspl4/objects.html#./objects:h10), is there a reason why this can't be a library with FFI?

harryprayiv:
Absolutely if you think that's a better course of action.

I'm using this backend to build super fast, lightweight executables. Purescript has become my mini Haskell and this back end compliments that intention perfectly.  


Ps. Whatever I can do to help, don't hesitate to ask. :)

